PR #72, opened 2022-06-13
url updated

VISHNUDAS-tunerlabs:


src/scripts/schedulerScript.js
```diff
@@ -19,7 +19,7 @@ const emailCronJobBeforeFifteenMinData = {
 	name: 'emailCronJobBeforeFifteenMin',
 	email: email,
 	request: {
-		url: 'http://localhost:3000/mentoring/v1/notifications/emailCronJobBeforeFifteenMin',
+		url: 'http://dev.elevate-apis.shikshalokam.org/mentoring/v1/notifications/emailCronJobBeforeFifteenMin',
```
rakeshSgr, line 22: base path should come from env

rakeshSgr: use localhost and port number, its already there in env

PR #89, opened 2022-07-13
Caching

ankitpws:
Implemented Caching Library and test cases for mentor profile is also updated

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -13,6 +13,7 @@ const path = require('path')
 const i18next = require('i18next')
 const Backend = require('i18next-fs-backend')
 const middleware = require('i18next-http-middleware')
+const { InternalCache, RedisHelper } = require('elevate-node-cache')
 
 let environmentData = require('./envVariables')()
 
@@ -23,6 +24,8 @@ if (!environmentData.success) {
 
 require('@configs')
 
+InternalCache.init(process.env.INTERNAL_CACHE_EXP_TIME)
+RedisHelper.init(process.env.REDIS_CACHE_EXP_TIME)
 i18next
 	.use(Backend)
 	.use(middleware.LanguageDetector)
--- a/src/app.js
+++ b/src/app.js
@@ -13,8 +13,6 @@ const path = require('path')
 const i18next = require('i18next')
 const Backend = require('i18next-fs-backend')
 const middleware = require('i18next-http-middleware')
-const { InternalCache, RedisHelper } = require('elevate-node-cache')
-
 let environmentData = require('./envVariables')()
 
 if (!environmentData.success) {
@@ -24,8 +22,6 @@ if (!environmentData.success) {
 
 require('@configs')
 
-InternalCache.init(process.env.INTERNAL_CACHE_EXP_TIME)
-RedisHelper.init(process.env.REDIS_CACHE_EXP_TIME)
 i18next
 	.use(Backend)
 	.use(middleware.LanguageDetector)
--- a/src/configs/index.js
+++ b/src/configs/index.js
@@ -9,4 +9,4 @@ require('./mongodb')()
 
 require('./kafka')()
 
-require('@ankitpws/caching-library').RedisConfig()
+require('elevate-node-cache').RedisConfig()
--- a/src/configs/index.js
+++ b/src/configs/index.js
@@ -9,4 +9,5 @@ require('./mongodb')()
 
 require('./kafka')()
 
-require('elevate-node-cache').RedisConfig()
+const { RedisConfig } = require('elevate-node-cache')
+RedisConfig.config(process.env.REDIS_HOST)
--- a/src/configs/index.js
+++ b/src/configs/index.js
@@ -8,3 +8,5 @@
 require('./mongodb')()
 
 require('./kafka')()
+
+require('@ankitpws/caching-library').RedisConfig()
--- a/src/configs/index.js
+++ b/src/configs/index.js
@@ -9,7 +9,4 @@ require('./mongodb')()
 
 require('./kafka')()
 
-const { RedisConfig, InternalCache, RedisHelper } = require('elevate-node-cache')
-RedisConfig.config(process.env.REDIS_HOST)
-InternalCache.init(process.env.INTERNAL_CACHE_EXP_TIME)
-RedisHelper.init(process.env.REDIS_CACHE_EXP_TIME)
+require('./cache')()
--- a/src/configs/index.js
+++ b/src/configs/index.js
@@ -9,5 +9,7 @@ require('./mongodb')()
 
 require('./kafka')()
 
-const { RedisConfig } = require('elevate-node-cache')
+const { RedisConfig, InternalCache, RedisHelper } = require('elevate-node-cache')
 RedisConfig.config(process.env.REDIS_HOST)
+InternalCache.init(process.env.INTERNAL_CACHE_EXP_TIME)
+RedisHelper.init(process.env.REDIS_CACHE_EXP_TIME)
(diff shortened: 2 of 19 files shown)


src/configs/index.js
```diff
@@ -8,3 +8,5 @@
 require('./mongodb')()
 
 require('./kafka')()
+
+require('@ankitpws/caching-library').RedisConfig()
```
rakeshSgr, line 12: @ankitpws keep the name as node-cache or node-caching, it shouldn't include any user name.

src/services/helper/entity.js
```diff
@@ -28,6 +28,8 @@ module.exports = class EntityHelper {
 				})
 			}
 			await entitiesData.createEntity(bodyData)
+			const key = 'mentoring_entity_' + bodyData.type
```
rakeshSgr, line 31: @ankitpws can we change this to _id_entity_type ? or any better way.

src/services/helper/entity.js
```diff
@@ -65,6 +67,16 @@ module.exports = class EntityHelper {
 					responseCode: 'CLIENT_ERROR',
 				})
 			}
+			console.log(_id)
```
rakeshSgr, line 70: @ankitpws console.log not required

src/services/helper/form.js
```diff
@@ -83,13 +86,20 @@ module.exports = class FormsHelper {
...
 	static async read(bodyData) {
 		try {
-			const form = await formsData.findOneForm(
-				bodyData.type,
-				bodyData.subType,
-				bodyData.action,
-				bodyData.ver,
-				bodyData.templateName
-			)
+			const key = bodyData.type + bodyData.subType + bodyData.action + bodyData.ver + bodyData.templateName
```
rakeshSgr, line 89: @ankitpws getting key should be from function. we should not construct in all the places. and set form type to unique in model.

src/services/helper/mentors.js
```diff
@@ -12,6 +12,7 @@ const apiEndpoints = require('@constants/endpoints')
 const apiBaseUrl = process.env.USER_SERIVCE_HOST + process.env.USER_SERIVCE_BASE_URL
 const request = require('@generics/requests')
 const sessionAttendees = require('@db/sessionAttendees/queries')
+const { RedisHelper } = require('@ankitpws/caching-library')
```
rakeshSgr, line 15: @ankitpws using config we should be able to set the cache expiry time separately for redies and internal cache

src/services/helper/mentors.js
```diff
@@ -76,28 +77,42 @@ module.exports = class MentorsHelper {
...
-			})
-		} else {
-			return common.failureResponse({
-				statusCode: httpStatusCode.bad_request,
-				message: 'MENTORS_NOT_FOUND',
-				responseCode: 'CLIENT_ERROR',
-			})
+		try {
+			let mentorsDetails = {}
+			if (await RedisHelper.getKey(id)) {
```
rakeshSgr, line 82: @ankitpws for checking the key you are getting cache data and again calling in the set line to send the, change this.

rakeshSgr: @ankitpws profile caching should be done in user service or here ?

src/configs/index.js
ankitpws: @rakeshSgr there are a lot of packages available on npm and this name is already taken so we can not name package as node-cache or node-caching instead we can make @sl/node-cache then we can publish package

src/services/helper/entity.js
ankitpws: @rakeshSgr we are not receiving id in API call so we can make id that's why i have used mentoring because in users module we have entity service

src/services/helper/entity.js
```diff
@@ -28,6 +28,8 @@ module.exports = class EntityHelper {
 				})
 			}
 			await entitiesData.createEntity(bodyData)
+			const key = 'mentoring_entity_' + bodyData.type
+			await InternalCache.delKey(key)
```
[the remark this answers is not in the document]
ankitpws: @rakeshSgr  when an entity is updated then that has to be cleared from the cache so that the updated one can take his place.

src/services/helper/entity.js
```diff
@@ -65,6 +67,16 @@ module.exports = class EntityHelper {
...
 			}
+			console.log(_id)
+			let key = ''
+			if (bodyData.type) {
+				key = 'mentoring_entity_' + bodyData.type
+				await InternalCache.delKey(key)
+			} else {
+				const entities = await entitiesData.findOne(_id)
+				key = 'mentoring_entity_' + entities.type
+				await InternalCache.delKey(key)
```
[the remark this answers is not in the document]
ankitpws: @rakeshSgr we then have to write list function query again in update and create to avoid that I am deleting key and when list API is called it will automatically cache the data.

src/db/forms/queries.js
```diff
@@ -25,6 +25,7 @@ module.exports = class FormsData {
 		return new Promise(async (resolve, reject) => {
 			try {
 				const formData = await Forms.findOne(filter, projection)
+				console.log(formData)
```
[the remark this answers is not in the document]
ankitpws: @rakeshSgr Fixed

src/services/helper/entity.js
ankitpws: @rakeshSgr Fixed

src/constants/common.js
```diff
@@ -5,13 +5,24 @@
  * Description : All commonly used constants through out the service
  */
 
-const successResponse = ({ statusCode = 500, responseCode = 'OK', message, result = [], meta = {} }) => {
+const FormsData = require('@db/forms/queries')
+const { InternalCache } = require('elevate-node-cache')
```
rakeshSgr, line 9: @ankitpws we should be able to set the expiry time while requiring the package.
And move InternalCache and Redis to utils, you can write multiple functions to do the job.

src/app.js
```diff
@@ -23,6 +24,8 @@ if (!environmentData.success) {
 
 require('@configs')
 
+InternalCache.init(process.env.INTERNAL_CACHE_EXP_TIME)
```
rakeshSgr, line 27: @ankitpws why we need here ?

src/app.js
```diff
@@ -13,6 +13,7 @@ const path = require('path')
 const i18next = require('i18next')
 const Backend = require('i18next-fs-backend')
 const middleware = require('i18next-http-middleware')
+const { InternalCache, RedisHelper } = require('elevate-node-cache')
```
rakeshSgr, line 16: @ankitpws what is the use of including in app.js?

src/services/helper/entity.js
```diff
@@ -1,10 +1,11 @@
 // Dependencies
 const ObjectId = require('mongoose').Types.ObjectId
-const utilsHelper = require('@generics/utils')
+
 const httpStatusCode = require('@generics/http-status')
 const common = require('@constants/common')
 const entitiesData = require('@db/entities/query')
-
+const { InternalCache } = require('elevate-node-cache')
```
rakeshSgr, line 7: @ankitpws this is not require right ?

src/services/helper/sessions.js
```diff
@@ -432,6 +432,7 @@ module.exports = class SessionsHelper {
 				)
 			}
 
+			console.log(sessionDetails)
```
rakeshSgr, line 435: @ankitpws not required

ankitpws: @rakeshSgr fixed

src/app.js
ankitpws: @rakeshSgr fixed

src/app.js
ankitpws: @rakeshSgr fixed

PR #99, opened 2022-07-25
Migrations

ankitpws:
Added migration module and script


src/module/migrations/Readme.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Check migrations
+
+    npm run elevate-migrations s
+
+## Create migrations
+
+    npm run elevate-migrations create <migration-name>
+
+## Run migrations
```
rakeshSgr, line 11: @ankitpws mention how to down the migration the migration also

rakeshSgr: @ankitpws add these data in root readme also

src/migrations/20220802074933-session_form.js
```diff
@@ -0,0 +1,128 @@
...
+								label: 'Session Title',
+								value: '',
+								class: 'ion-margin',
+								type: 'text',
+								position: 'floating',
+								validators: { required: true },
+							},
+							{
+								name: 'description',
+								label: 'Description',
```
rakeshSgr, line 37: @ankitpws value , class,position and validators are the same for all fields, no need to mention in the all the places, clear the code

PR #103, opened 2022-07-26
Fixed Mentor Name

ankitpws:
Fixed Mentor Name

### DIFF
--- a/src/services/helper/sessions.js
+++ b/src/services/helper/sessions.js
@@ -15,6 +15,7 @@ const request = require('request')
 const bigBlueButton = require('./bigBlueButton')
 const userProfile = require('./userProfile')
 const utils = require('@generics/utils')
+const UserProfileHelper = require('./userProfile')
 
 module.exports = class SessionsHelper {
 	/**
@@ -43,6 +44,9 @@ module.exports = class SessionsHelper {
 			if (bodyData.endDate) {
 				bodyData['endDateUtc'] = moment.unix(bodyData.endDate).utc().format(common.UTC_DATE_TIME_FORMAT)
 			}
+
+			const mentorName = await UserProfileHelper.details('', loggedInUserId)
+			req.body.mentorName = mentorName.data.result.name
 
 			let elapsedMinutes = moment(bodyData.endDateUtc).diff(bodyData.startDateUtc, 'minutes')
 
@@ -340,6 +344,8 @@ module.exports = class SessionsHelper {
 				})
 				sessionDetails.image = await Promise.all(sessionDetails.image)
 			}
+			const mentorName = await UserProfileHelper.details('', sessionDetails.userId)
+			sessionDetails.mentorName = mentorName.data.result.name
 
 			return common.successResponse({
 				statusCode: httpStatusCode.created,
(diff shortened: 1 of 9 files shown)


src/services/helper/sessions.js
```diff
@@ -44,6 +45,9 @@ module.exports = class SessionsHelper {
 				bodyData['endDateUtc'] = moment.unix(bodyData.endDate).utc().format(common.UTC_DATE_TIME_FORMAT)
 			}
 
+			const mentorName = await UserProfileHelper.details('', loggedInUserId)
```
rakeshSgr, line 48: @ankitpws this is not required, and don't store mentor name in DB.

In response only we need to send the result mentor name

src/services/helper/sessions.js
```diff
@@ -426,14 +432,23 @@ module.exports = class SessionsHelper {
 						}
 					})
 				)
+
+				await Promise.all(
+					sessionDetails[0].data.map(async (session) => {
+						const mentorName = await UserProfileHelper.details('', session.userId)
```
rakeshSgr, line 438: @ankitpws  get all the users data in one API

src/services/helper/sessions.js
```diff
@@ -426,14 +432,23 @@ module.exports = class SessionsHelper {
...
+						session.mentorName = mentorName.data.result.name
+					})
+				)
 			}
+			console.log(sessionDetails[0].data)
```
rakeshSgr, line 443: @ankitpws remove this

src/services/helper/sessions.js
```diff
@@ -550,6 +568,10 @@ module.exports = class SessionsHelper {
 					responseCode: 'CLIENT_ERROR',
 				})
 			}
+
```
rakeshSgr, line 571: @ankitpws check with frontent team, they require mentor name when they do enroll and unenroll the session, If require then only add it.

rakeshSgr: @ankitpws sending a email we may require.

ankitpws: for Email only we require mentor name @rakeshSgr

src/services/helper/sessions.js
ankitpws: @rakeshSgr there is no API which will give me all the mentor details

rakeshSgr: check this getAllAccountsDetail

src/services/helper/sessions.js
```diff
@@ -426,6 +430,19 @@ module.exports = class SessionsHelper {
 						}
 					})
 				)
+				const userIds = sessionDetails[0].data
+					.map((item) => item.userId.toString())
+					.filter((value, index, self) => self.indexOf(value) === index)
+
+				let mentorDetails = await UserProfileHelper.getAllMentorsDetail(userIds)
```
rakeshSgr, line 437: @ankitpws change to getListOfUserDetails in enhancement

rakeshSgr: @ankitpws we don't require all the fields, can we pass only the required fields from here, take it enahncement

src/services/helper/sessions.js
```diff
@@ -426,6 +430,19 @@ module.exports = class SessionsHelper {
...
+				mentorDetails = mentorDetails.result
+
+				for (let i = 0; i < sessionDetails[0].data.length; i++) {
+					let mentorIndex = mentorDetails.findIndex(
```
rakeshSgr, line 441: @ankitpws improve this logic, if store as object, then you can easily access.
sample : {
   "1214123" :{ ...data }
}

PR #161, opened 2022-11-17
added lean plugin to all services

ankitpws:
added lean plugin to all services


src/services/helper/feedback.js
```diff
@@ -201,6 +201,7 @@ module.exports = class MenteesHelper {
 					responseCode: 'CLIENT_ERROR',
 				})
 			}
+			console.log(sessionInfo)
```
rakeshSgr, line 204: @ankitpws remove

rakeshSgr: console log

PR #182, opened 2023-01-24
Enroll unenroll changes

nevil-mathew:


src/integration-tests/sessions/responseSchema.js
```diff
@@ -607,18 +607,59 @@ const enrollSchema = {
...
+				'link',
+				'skippedFeedback',
+				'_id',
+				'feedbacks',
+				'updatedAt',
+				'createdAt',
+				'__v',
+			],
+		},
+		meta: {
```
rakeshSgr, line 661: @nevil-mathew Are we using this key now ?

PR #184, opened 2023-01-30
Session update changes

nevil-mathew:


src/services/helper/sessions.js
```diff
@@ -292,6 +295,7 @@ module.exports = class SessionsHelper {
 				message: message,
 			})
 		} catch (error) {
+			console.log(error)
```
rakeshSgr, line 298: @nevil-mathew use logger instead of console.log

nevil-mathew: removed

PR #250, opened 2023-07-31
Develop 2.5

joffinjoy:


src/database/queries/questions.js
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	static async updateOneQuestion(filter, update, options = {}) {
+		try {
+			const [rowsAffected] = await Question.update(update, {
+				where: filter,
+				...options,
+			})
+
+			return rowsAffected > 0 ? 'QUESTION_UPDATED' : 'QUESTION_NOT_FOUND'
```
joffinjoy, line 46: Note to Priyanka... If not the affectedRowsCount, we can use this pattern. But let's keep on pattern across services.

src/services/helper/form.js
```diff
@@ -51,20 +52,18 @@ module.exports = class FormsHelper {
...
 			let filter = {}
-			if (ObjectId.isValid(id)) {
+			if (id) {
 				filter = {
-					_id: ObjectId(id),
+					id: id,
 				}
 			} else {
 				filter = {
 					type: bodyData.type,
```
joffinjoy, line 61: Note to self: Same pattern in user service. 

To dev: Since we are using form versions. Shouldn't it be part of the filter to identify which is the form that is actually getting updated in-case there is no id?

src/services/helper/form.js
```diff
@@ -86,6 +85,13 @@ module.exports = class FormsHelper {
 				message: 'FORM_UPDATED_SUCCESSFULLY',
 			})
 		} catch (error) {
+			if (error instanceof UniqueConstraintError) {
```
joffinjoy, line 88: Check how is this contrainst handled in user service

src/services/helper/form.js
joffinjoy: Comment [Nevil]: The previous version isn't stored but gets overwritten.

PR #258, opened 2023-08-14
updated code to reflect PR comments

nevil-mathew:


src/database/models/entitytype.js
```diff
@@ -14,11 +14,8 @@ module.exports = (sequelize, DataTypes) => {
 			status: { type: DataTypes.STRING, allowNull: false, defaultValue: 'ACTIVE' },
 			created_by: { type: DataTypes.INTEGER, allowNull: false },
 			updated_by: { type: DataTypes.INTEGER, allowNull: false },
-			allow_filtering: {
-				type: DataTypes.BOOLEAN,
-				allowNull: false,
-			},
-			data_type: { type: DataTypes.STRING, allowNull: false },
+			allow_filtering: { type: DataTypes.BOOLEAN, allowNull: false, defaultValue: 'TRUE' },
```
joffinjoy, line 17: allow_filtering should be false by default and check if the defaultValue should be true or "TRUE" since this is a Boolean data type.

src/database/queries/mentorextention.js
```diff
@@ -9,23 +9,17 @@ module.exports = class MentorExtensionQueries {
 		}
 	}
```
joffinjoy, line 11: rename this file to mentorExtension.js

PR #265, opened 2023-09-07
Add mentor validation

nevil-mathew:


src/constants/common.js
```diff
@@ -82,4 +83,8 @@ module.exports = {
 		'/mentors/upcomingSessions/',
 		'/issues/create',
 	],
+	MENTOR_ROLE: 'mentor',
```
joffinjoy, line 86: This key is repeated. Remove one

src/database/models/sessions.js
```diff
@@ -79,17 +82,20 @@ module.exports = (sequelize, DataTypes) => {
 			is_feedback_skipped: {
 				type: DataTypes.BOOLEAN,
 				allowNull: false,
+				defaultValue: false,
 			},
 			mentee_feedback_question_set: {
 				type: DataTypes.STRING,
-				allowNull: false,
+				allowNull: true,
+				defaultValue: 'MENTEE_QS1',
```
joffinjoy, line 90: Implementation wise it's okey. Pattern wise, might not be the best fit.

src/database/queries/sessionAttendees.js
```diff
@@ -0,0 +1,124 @@
...
+
+exports.updateOne = async (filter, update, options = {}) => {
+	try {
+		const [rowsAffected] = await SessionAttendee.update(update, {
+			where: filter,
+			...options,
+			individualHooks: true, // Pass 'individualHooks: true' option to ensure proper triggering of 'beforeUpdate' hook.
+		})
+
+		if (rowsAffected > 0) {
```
joffinjoy, line 34: We discarded this pattern right? Always return the raw database object and handle it at the service or caller layer

src/database/queries/sessionAttendees.js
```diff
@@ -0,0 +1,124 @@
...
+}
+
+exports.unEnrollAllAttendeesOfSessions = async (sessionIds) => {
+	try {
+		const destroyedCount = await SessionAttendee.destroy({
+			where: {
+				session_id: { [Op.in]: sessionIds },
+			},
+		})
+		await SessionEnrollment.destroy({
```
joffinjoy, line 78: Probably better to have a function SessionEnrollment's query file and call it here.

src/database/queries/sessions.js
```diff
@@ -0,0 +1,295 @@
...
+					model: Session,
+					where: {
+						deleted: false,
+						[Op.or]: [{ startDate: { [Op.gt]: currentEpochTime } }, { status: common.PUBLISHED_STATUS }],
+					},
+					attributes: ['id', 'title'],
+				},
+			],
+		}) */
+		const foundSessionOwnerships = await SessionOwnership.findAll({
```
joffinjoy, line 179: [DISCUSS] If model functions of another table be invovked directly in this file's function?

src/generics/utils.js
```diff
@@ -161,6 +162,16 @@ function redisDel(key) {
 const capitalize = (str) => {
 	return startCase(str)
 }
+const isAMentor = (roles) => {
```
joffinjoy, line 165: Since this is exported as a function, use this everywhere.

src/services/helper/sessions.js
```diff
@@ -77,21 +77,83 @@ module.exports = class SessionsHelper {
...
-			bodyData.meetingInfo = {
+
+			//validate entities
+			if (
+				bodyData.hasOwnProperty(common.MEDIUM) ||
+				bodyData.hasOwnProperty(common.RECOMMENDED_FOR) ||
+				bodyData.hasOwnProperty(common.CATEGORIES)
+			) {
+				let values = []
+				if (bodyData.hasOwnProperty(common.MEDIUM)) values.push(common.MEDIUM)
```
joffinjoy, line 88: Better structing? Multiple same if checks.

src/database/models/sessions.js
joffinjoy: Get this value from process.env

PR #269, opened 2023-10-01
KATHA-2339: Changes Related To Job Scheduling

VISHNUDAS-tunerlabs:
The PR includes changes related to **KATHA-2339.** You can find more details about this task at [Task-2339](https://katha.shikshalokam.org/task-view-2339.html)."

**### Brief of Changes:**

- Added a new API endpoint to the notification controller for elevate-scheduler service to call.
- Safely removed three old API endpoints.
- Implemented logic to call mentor and mentee notifications.
- Added request module and endpoints for mentor service's communication with the scheduler service.
- Made environmental key changes.

**### .env Changes Required for Task 2339:**

Two new environment keys need to be added (sample provided below):
**SCHEDULER_SERVICE_HOST="http://localhost:4000/"
SCHEDULER_SERVICE_BASE_URL= '/scheduler/'**

Reminder to Check: After these changes, we may no longer need migration scripts scheduler.js and bull.js. If so, we can also remove one existing .env variable, SCHEDULER_SERVICE_URL."

### DIFF
--- a/src/constants/common.js
+++ b/src/constants/common.js
@@ -87,4 +87,27 @@ module.exports = {
 	MEDIUM: 'medium',
 	RECOMMENDED_FOR: 'recommended_for',
 	CATEGORIES: 'categories',
+	jobsToCreate: [
+		{
+			jobId: 'mentoring_session_one_hour_',
+			jobName: 'notificationBeforeAnHour',
+			emailTemplate: 'mentor_one_hour_before_session_reminder',
+		},
+		{
+			jobId: 'mentoring_session_one_day_',
+			jobName: 'notificationBeforeOneDay',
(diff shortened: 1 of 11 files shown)


src/constants/common.js
```diff
@@ -87,4 +87,27 @@ module.exports = {
 	MEDIUM: 'medium',
 	RECOMMENDED_FOR: 'recommended_for',
 	CATEGORIES: 'categories',
+	jobsToCreate: [
```
joffinjoy, line 90: This basically removes the one time creation of jobs in scheduler service completely right?

VISHNUDAS-tunerlabs: @joffinjoy Yes.

PR #289, opened 2023-10-12
Project restructuring

joffinjoy:


src/controllers/v1/mentees.js
```diff
@@ -18,13 +18,13 @@ module.exports = class Mentees {
 	 * @param {String} req.decodedToken.id - User Id.
 	 * @returns {JSON} - Mentee profile details
 	 */
-	async profile(req) {
+	/* async profile(req) {
```
VISHNUDAS-tunerlabs, line 21: @joffinjoy we don't need this api? In develop-2.5 This code is not commented

joffinjoy: Yeah! Not needed.

PR #443, opened 2023-12-27
Query optimisation

nevil-mathew:


src/services/entity-type.js
```diff
@@ -89,7 +89,15 @@ module.exports = class EntityHelper {
 		try {
```
joffinjoy, line 89: Modify the variable name to orgId

src/services/questionsSet.js
```diff
@@ -107,11 +107,14 @@ module.exports = class questionsSetHelper {
 	 * @returns {JSON} - Read question set.
 	 */
 
-	static async read(questionsSetId) {
+	static async read(questionsSetId, questionSetCode) {
 		try {
-			const filter = {
+			let filter = {
 				id: questionsSetId,
 			}
```
joffinjoy, line 114: const is fine

src/database/queries/entityType.js
```diff
@@ -99,11 +95,12 @@ module.exports = class UserEntityData {
 		}
 	} */
 
-	static async updateOneEntityType(id, update, options = {}) {
+	static async updateOneEntityType(id, organizationId, update, options = {}) {
```
joffinjoy, line 98: make this orgId

PR #446, opened 2023-12-29
[Task#2448] Changes required to enable mentor listing cards

nevil-mathew:


src/services/sessions.js
```diff
@@ -82,6 +84,17 @@ module.exports = class SessionsHelper {
 				})
 			}
 
+			if (!bodyData.mentor_id) {
```
aks30, line 87: @nevil-mathew we should validate if the mentor ids are valid and as per policy, also type and mentor_id are exclusive, meaning type can be pvt/public. mentor_id if given in req body we should take that else use the logged in user's id.

src/services/sessions.js
```diff
@@ -234,7 +249,13 @@ module.exports = class SessionsHelper {
 					responseCode: 'CLIENT_ERROR',
 				})
 			}
-
+			if (!sessionDetail.created_by !== userId) {
+				return common.failureResponse({
+					message: 'CANNOT_EDIT_DELETE_SESSION_CREATED_BY_MANAGER',
```
aks30, line 254: @nevil-mathew Let this be a generic message, cannot edit the session, manager or not we need not tell

src/locales/en.json
```diff
@@ -135,5 +135,7 @@
 	"MODULES_NOT_DELETED": "Module are not deleted",
 	"MODULES_DELETED_SUCCESSFULLY": "Module deleted successfully",
 	"MODULES_HAS_EMPTY_LIST": "Module are empty for searchlist",
-	"MODULES_FETCHED_SUCCESSFULLY": "Module fetched successfully"
+	"MODULES_FETCHED_SUCCESSFULLY": "Module fetched successfully",
+	"PRIVATE_TYPE_NOT_ALLOWED": "Sorry, the specified type is not accessible for this session. Please check the type permissions.",
+	"CANNOT_EDIT_DELETE_SESSION_CREATED_BY_MANAGER": "Sorry, you are not authorized to edit or delete sessions created by the session manager."
```
aks30, line 140: @nevil-mathew message can be - "Sorry, you are not authorized to edit or delete this session"

aks30 (changes requested):
Reviewed on 2 Jan

src/database/migrations/20231228184838-add-columns-to-sessions.js
```diff
@@ -0,0 +1,25 @@
...
+	up: async (queryInterface, Sequelize) => {
+		await queryInterface.addColumn('sessions', 'created_by', {
+			type: Sequelize.INTEGER,
+			allowNull: true,
+		})
+		await queryInterface.addColumn('sessions', 'updated_by', {
+			type: Sequelize.INTEGER,
+			allowNull: true,
+		})
+		await queryInterface.addColumn('sessions', 'type', {
```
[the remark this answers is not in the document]
nevil-mathew: We cannot set allowNull as false, it will cause issues if data exists.
```
Sequelize CLI [Node: 16.16.0, CLI: 6.6.1, ORM: 6.32.0]

Loaded configuration file "configs/postgres.js".
Using environment "development".
== 20231228184838-add-columns-to-sessions: migrating =======

ERROR: column "updated_by" of relation "sessions" contains null values
```
So I've split migration into a two step process.

PR #449, opened 2024-01-03
org filter list

priyanka-TL:


src/controllers/v1/profile.js
```diff
@@ -97,6 +97,27 @@ module.exports = class Mentees {
...
+
+	async filterList(req) {
+		try {
+			const filterList = await menteesService.getFilterList(
+				req.query.entity_types ? req.query.entity_types : '',
+				req.decodedToken
+			)
+			return filterList
+		} catch (error) {
+			return error
```
joffinjoy, line 117: Do not return the error here. Console.log it and throw again. 

Otherwise, remove the try catch block altogether. A catch block that just throws an execption isn't doing anything.

src/services/mentees.js
```diff
@@ -779,4 +780,207 @@ module.exports = class MenteesHelper {
...
+
+			const menteeExtension = await menteeQueries.getMenteeExtension(userId, [
+				'external_mentor_visibility',
+				'organization_id',
+			])
+
+			if (menteeExtension.external_mentor_visibility === common.CURRENT) {
+				organization_ids.push(menteeExtension.organization_id)
+			} else if (menteeExtension.external_mentor_visibility === common.ASSOCIATED) {
+				let userOrgDetails = await userRequests.fetchDefaultOrgDetails(menteeExtension.organization_id)
```
joffinjoy, line 863: Though functionally this is fine, this function should be made generic since here we are not fetching the default organisation details. Leave unchanged for now.

src/services/mentees.js
```diff
@@ -779,4 +780,207 @@ module.exports = class MenteesHelper {
...
+				if (
+					userOrgDetails.success &&
+					userOrgDetails.data &&
+					userOrgDetails.data.result &&
+					userOrgDetails.data.result.related_orgs &&
+					userOrgDetails.data.result.related_orgs.length > 0
+				) {
+					organization_ids.push(...userOrgDetails.data.result.related_orgs)
```
joffinjoy, line 871: Do we not need mentee's own organisation id here? This only gets associated org ids right?

src/services/mentees.js
```diff
@@ -779,4 +780,207 @@ module.exports = class MenteesHelper {
...
+			const organizations = await this.getOrganizationIdBasedOnPolicy(tokenInformation.id)
+			if (organizations.success && organizations.result.length > 0) {
+				organization_ids = [...organizations.result]
+
+				const defaultOrgId = await getDefaultOrgId()
+				if (defaultOrgId && !organization_ids.includes(defaultOrgId)) {
+					organization_ids.push(defaultOrgId)
+				}
+
+				let findOrganizationAndEntities = false
```
joffinjoy, line 809: const findOrganizationAndEntities = organization_ids.length > 0 ? true: false;

Should be cleaner and equivalent

src/services/mentees.js
```diff
@@ -779,4 +780,207 @@ module.exports = class MenteesHelper {
...
+				if (organization_ids.length > 0) {
+					findOrganizationAndEntities = true
+				}
+
+				if (findOrganizationAndEntities) {
```
joffinjoy, line 814: Since this is the only place this variable is used, you can simply check for organization_ids.length >0 without using this new variable.

src/services/mentees.js
```diff
@@ -779,4 +780,207 @@ module.exports = class MenteesHelper {
...
+			if (entity_types) {
+				entityTypes = entity_types.split(',')
+				filter.value = {
+					[Op.in]: entityTypes,
+				}
+			}
+
+			//fetch entity types and entities
+			let entityTypesWithEntities = await entityTypeQueries.findUserEntityTypesAndEntities(filter)
+			entityTypesWithEntities = JSON.parse(JSON.stringify(entityTypesWithEntities))
```
joffinjoy, line 927: Why is this parsed like this?

src/database/queries/organisationExtension.js
```diff
@@ -61,4 +61,17 @@ module.exports = class OrganizationExtensionQueries {
 			throw new Error(`Error finding/inserting organisation extension: ${error.message}`)
 		}
 	}
+
+	static async findAll(filter, options = {}) {
+		try {
+			const orgPolicies = await OrganizationExtension.findAll({
```
joffinjoy, line 67: rename this to orgExtensions

src/services/mentees.js
```diff
@@ -781,7 +782,146 @@ module.exports = class MenteesHelper {
...
+				organization_ids.push(menteeExtension.organization_id)
+			} else if (menteeExtension.external_mentor_visibility === common.ASSOCIATED) {
+				let userOrgDetails = await userRequests.fetchDefaultOrgDetails(menteeExtension.organization_id)
+				if (userOrgDetails.success && userOrgDetails.data?.result?.related_orgs?.length > 0) {
+					organization_ids.push(...userOrgDetails.data.result.related_orgs, menteeExtension.organization_id)
+				}
+			} else if (menteeExtension.external_mentor_visibility === common.ALL) {
+				const organizationExtension = await organisationExtensionQueries.findAll(
+					{
+						external_mentor_visibility_policy: common.ALL,
```
joffinjoy, line 860: External_mentor_visibility_policy seems irrelvant here. Please remove

src/services/mentees.js
```diff
@@ -781,7 +782,146 @@ module.exports = class MenteesHelper {
...
+						mentor_visibility_policy: {
```
joffinjoy, line 861: Modify this query such that all all organisations with mentor_visibility_policy as ALL is fetched + all associated organisations of the current org is also fetched.

src/services/mentees.js
```diff
@@ -781,7 +782,146 @@ module.exports = class MenteesHelper {
...
+						},
+					},
+					{
+						attributes: ['organization_id'],
+					}
+				)
+
+				if (organizationExtension) {
+					const organizationIds = organizationExtension.map((orgExt) => orgExt.organization_id)
+					if (organizationIds.length > 0) {
```
joffinjoy, line 872: This if check might not be required, since spreading an empty array might result in another empty array.

src/services/mentees.js
```diff
@@ -781,7 +782,146 @@ module.exports = class MenteesHelper {
...
+			}
+
+			let organization_ids = []
+			const organizations = await this.getOrganizationIdBasedOnPolicy(tokenInformation.id)
+			if (organizations.success && organizations.result.length > 0) {
+				organization_ids = [...organizations.result]
+
+				if (organization_ids.length > 0) {
+					//get organization list
+					const organizations = await userRequests.listOrganization(organization_ids)
```
joffinjoy, line 806: Suggest that you modify the name here since it is shadowing the same variable name from parent scope.

src/requests/user.js
```diff
@@ -232,6 +232,38 @@ const search = function (userType, pageNo, pageSize, searchText, userServiceQuer
...
+				internal_access_token: process.env.INTERNAL_ACCESS_TOKEN,
+			},
+			form: {
+				organizationIds,
+			},
+		}
+
+		const orgListUrl = userBaseUrl + endpoints.ORGANIZATION_LIST
+		try {
+			request.get(orgListUrl, options, callback)
```
joffinjoy, line 250: Use the generic requests functions for making this change, instead of using request package.

PR #475, opened 2024-01-16
added enrolledMentees list api

nevil-mathew:


src/services/mentees.js
```diff
@@ -1042,6 +1056,18 @@ module.exports = class MenteesHelper {
 				additionalProjectionString,
 				false
 			)
+			mentorExtensionDetails = await mentorQueries.getMentorsByUserIdsFromView(
```
rakeshSgr, line 1059: @nevil-mathew Why was this added multiple times?

nevil-mathew: That is for the pagination logic. This is the existing flow. @rakeshSgr

src/services/sessions.js
```diff
@@ -1698,4 +1698,137 @@ module.exports = class SessionsHelper {
...
+				'last_logged_in_at',
+				'has_accepted_terms_and_conditions',
+				'languages',
+				'preferred_language',
+				'custom_entity_text',
+			]
+
+			const cleanedAttendeesAccounts = mergedUserArray.map((user) => {
+				propertiesToDelete.forEach((property) => {
+					delete user[property]
```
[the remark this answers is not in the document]
nevil-mathew: We can have that, but few of the keys come from the user service. That's why I went with this flow.

PR #490, opened 2024-01-19
mentees mentors count issue

adithyadinesh0412:


src/services/mentees.js
```diff
@@ -1107,9 +1108,6 @@ module.exports = class MenteesHelper {
 				})
 				.filter((value) => value !== null)
 
-			// update count after filters
-			userDetails.data.result.count = userDetails.data.result.count
```
rakeshSgr, line 1111: @adithyadinesh0412 why this line removed

adithyadinesh0412: userDetails already have count in it. And the count is not modified. We are using the count from userDetails.
It will be returned in the response.

src/middlewares/pagination.js
```diff
@@ -16,10 +16,10 @@ module.exports = (req, res, next) => {
...
 	req.searchText = req.query.search && req.query.search != '' ? decodeURI(req.query.search) : ''
-	// if (req.searchText != '') {
-	// 	let buff = new Buffer.from(req.searchText, 'base64')
-	// 	req.searchText = buff.toString('ascii')
-	// }
+	if (req.searchText != '') {
+		let buff = new Buffer.from(req.searchText, 'base64')
+		req.searchText = buff.toString('ascii')
+	}
 	let x = containsSpecialChars(req.searchText)
```
joffinjoy, line 23: Remove this x as this is unused

joffinjoy: Remove this emoji also

PR #491, opened 2024-01-19
Entity type validation

nevil-mathew:


src/services/mentees.js
```diff
@@ -45,12 +45,14 @@ module.exports = class MenteesHelper {
...
 				responseCode: 'CLIENT_ERROR',
 			})
+		const userExtensionsModelName = await menteeQueries.getModelName()
 
 		let entityTypes = await entityTypeQueries.findUserEntityTypesAndEntities({
 			status: 'ACTIVE',
 			organization_id: {
 				[Op.in]: [orgId, defaultOrgId],
 			},
+			model_names: { [Op.overlap]: [userExtensionsModelName] },
```
joffinjoy, line 55: Why are we using overalp here? Is array to array comparision necessary here?

src/services/mentees.js
```diff
@@ -1011,12 +1019,19 @@ module.exports = class MenteesHelper {
 			}
 
 			const query = utils.processQueryParametersWithExclusions(queryParams)
+			const userExtensionModelName = await menteeQueries.getModelName()
+			const mentorExtensionModelName = await menteeQueries.getModelName()
 
 			let validationData = await entityTypeQueries.findAllEntityTypesAndEntities({
 				status: common.ACTIVE_STATUS,
+				model_names: { [Op.overlap]: [userExtensionModelName, mentorExtensionModelName] },
```
joffinjoy, line 1027: I think we should use overlap only when necessary as given here. Other instances can be without overlap, instead contains should suffice.

src/services/mentees.js
```diff
@@ -1011,12 +1019,19 @@ module.exports = class MenteesHelper {
...
 			})
 
-			let filteredQuery = utils.validateFilters(query, JSON.parse(JSON.stringify(validationData)), 'sessions')
+			let filteredQuery = utils.validateFilters(
+				query,
+				JSON.parse(JSON.stringify(validationData)),
+				userExtensionModelName
```
joffinjoy, line 1033: Earlier this was "sessions" right? Was that a mistake?

src/services/mentees.js
```diff
@@ -759,14 +766,15 @@ module.exports = class MenteesHelper {
...
 					responseCode: 'CLIENT_ERROR',
 				})
-
+			const userExtensionsModelName = await menteeQueries.getModelName()
 			const filter = {
 				status: 'ACTIVE',
 				organization_id: {
 					[Op.in]: [orgId, defaultOrgId],
 				},
+				model_names: { [Op.contains]: [userExtensionsModelName] },
```
joffinjoy, line 775: Make this single element search

PR #500, opened 2024-01-21
Org extension create

joffinjoy:


src/database/queries/organisationExtension.js
```diff
@@ -86,4 +86,13 @@ module.exports = class OrganizationExtensionQueries {
 			throw new Error(`Error fetching organisation extension: ${error.message}`)
 		}
 	}
+
+	static async create(data, options = {}) {
+		try {
+			const newOrgExtension = await OrganizationExtension.create(data, options)
```
aks30, line 92: @joffinjoy why is it not consistently  camel case OrganizationExtension => organizationExtension

aks30 (changes requested):
Reviewed on 22 Jan

src/services/organization.js
```diff
@@ -52,6 +53,45 @@ module.exports = class OrganizationService {
 			})
 		} catch (error) {
 			console.log(error)
+			throw error
+		}
+	}
+
+	static async createOrgExtension(eventBody) {
+		try {
+			const extensionData = {
```
rakeshSgr, line 62: @joffinjoy can we use the eventBody Object to upsert, instead of accessing and storing the fields

src/helpers/eventListnerRouter.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
+
+exports.eventListenerRouter = async (eventBody, { createFn = null, updateFn = null, deleteFn = null }) => {
+	try {
+		switch (eventBody.eventType) {
```
rakeshSgr, line 5: @joffinjoy for new event type we need to upset this ?

PR #503, opened 2024-01-22
removed all mongo related dependencies

nevil-mathew:


src/package.json
rakeshSgr, line 1: @nevil-mathew remove this @shelf/jest-mongodb in user service also

PR #504, opened 2024-01-22
Add EntityType/Entity role based api validations. Katha #2426

adithyadinesh0412:


src/database/migrations/20240122093255-add-role-permissions-entity.js
```diff
@@ -0,0 +1,132 @@
...
+			common.pagination.DEFAULT_PAGE_NO,
+			common.pagination.DEFAULT_PAGE_SIZE,
+			common.SEARCH
+		)
+		const allRoles = response.result.data
+		if (!allRoles || !Array.isArray(allRoles)) {
+			throw new Error('No roles found.')
+		}
+		const roleIds = allRoles.map((role) => role.id)
+		const titles = [common.SESSION_MANAGER_ROLE, common.ADMIN_ROLE, common.MENTOR_ROLE, common.ORG_ADMIN_ROLE]
```
rakeshSgr, line 23: @adithyadinesh0412 we should make this API should be access to all the roles

PR #511, opened 2024-01-24
env variable clean up

priyanka-TL:


dev-ops/integration_test.user.env
```diff
@@ -9,10 +9,6 @@ APPLICATION_ENV = development
 # Database connectivity url
 MONGODB_URL = mongodb://mongo:27017/elevate-mentoring
```
rakeshSgr, line 10: @priyanka-TL mongodb connection url not require

priyanka-TL: @rakeshSgr Nevil is working on clearing mongo task so i skipped it

PR #532, opened 2024-01-30
[Tech Debt] : responses Helper Added In The Service

VISHNUDAS-tunerlabs:


src/middlewares/authenticator.js
```diff
@@ -15,24 +15,25 @@ const rolePermissionMappingQueries = require('@database/queries/rolePermissionMa
 const permissionsQueries = require('@database/queries/permissions')
 
 module.exports = async function (req, res, next) {
+	const unAuthorizedResponse = common.failureResponse({
```
joffinjoy, line 18: This is unchanged. Probably missed this modification after a pull from develop. Always make sure that the new code being pulled is also compatible with new changes if you are doing the pull after the commits.

VISHNUDAS-tunerlabs: @joffinjoy This is missed out in the new changes. That is why I mentioned we have to re check  this after all changes merged to develop before deploying this change. What do you suggest?

joffinjoy: Just to a delta check when raising the PR. If there are new codes yet to be merged, check if it compatible with your changes.

PR #538, opened 2024-01-31
List RolePermissions API

sumanvpacewisdom:


src/controllers/v1/rolePermissionMapping.js
```diff
@@ -44,4 +44,21 @@ module.exports = class RolePermission {
...
+	 * list rolePermission.
+	 * @method
+	 * @name list
+	 * @param {Integer} req.decodedToken.roles - roles
+	 * @returns {JSON} - RolePermission list object.
+	 */
+
+	async list(req) {
+		try {
+			const roleIds = Array.isArray(req.decodedToken.roles) ? req.decodedToken.roles.map(({ id }) => id) : []
```
joffinjoy, line 58: You are doing an array check here. First verify if this roles will ever be anything other than an array. If it is always going to be an array, then you can remove this check and then just do a map on req.decodedToken.roles.

src/services/rolePermissionMapping.js
```diff
@@ -86,4 +86,50 @@ module.exports = class modulesHelper {
...
+
+	/**
+	 * list rolePermission.
+	 * @method
+	 * @name list
+	 * @param {Integer} roleId - role_id
+	 * @returns {JSON} - RolePermission list object.
+	 */
+
+	static async list(roleId) {
```
joffinjoy, line 98: Argument name should be in plural since you are passing an array of roles

src/services/rolePermissionMapping.js
```diff
@@ -86,4 +86,50 @@ module.exports = class modulesHelper {
...
+		try {
+			const filter = { role_id: roleId }
+			const attributes = ['module', 'request_type']
+			const permissionAndModules = await rolePermissionMappingQueries.find(filter, attributes)
```
joffinjoy, line 102: Rename this function to findAll

src/controllers/v1/rolePermissionMapping.js
sumanvpacewisdom: I verified it will work without array check , it comes in array format only.

src/services/rolePermissionMapping.js
sumanvpacewisdom: i have changed as roleIds

src/locales/en.json
```diff
@@ -151,5 +151,6 @@
 	"ROLE_PERMISSION_NOT_FOUND": "Permission for this role is empty",
 	"PERMISSION_DENIED": "You do not have the required permissions to access this resource. Please contact your administrator for assistance.",
 	"SESSION_ATTENDEES": "Mentee list fetched successfully",
+	"FETCHED_ROLE_PERMISSION_SUCCESSFULLY": "permissions list fetched successfully"
```
joffinjoy, line 154: missing comma

PR #539, opened 2024-02-01
[Bug - 1187]Round off Applied To Session Duration Passed to the Mail

VISHNUDAS-tunerlabs:
Fix :  Duration should be rounded off to specific time, for example(60.55555 indicates 61 and 60.22222 indicates 60)

### DIFF
--- a/src/services/sessions.js
+++ b/src/services/sessions.js
@@ -1239,7 +1239,7 @@ module.exports = class SessionsHelper {
 							unitOfTime: common.UNIT_OF_TIME,
 							startDate: utils.getTimeZone(session.start_date, common.dateFormat, session.time_zone),
 							startTime: utils.getTimeZone(session.start_date, common.timeFormat, session.time_zone),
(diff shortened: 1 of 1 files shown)


src/services/sessions.js
```diff
@@ -1238,7 +1239,7 @@ module.exports = class SessionsHelper {
 							unitOfTime: common.UNIT_OF_TIME,
 							startDate: utils.getTimeZone(session.start_date, common.dateFormat, session.time_zone),
 							startTime: utils.getTimeZone(session.start_date, common.timeFormat, session.time_zone),
-							sessionDuration: sessionDuration,
+							sessionDuration: Math.duration(sessionDuration),
```
rakeshSgr, line 1242: @VISHNUDAS-tunerlabs is this correct ?

VISHNUDAS-tunerlabs: @rakeshSgr corrected

PR #592, opened 2024-02-15
Katha #1205 - mentee mentor list fix

adithyadinesh0412:


src/services/mentees.js
```diff
@@ -1211,11 +1199,17 @@ module.exports = class MenteesHelper {
 			throw error
 		}
 	}
-	static async filterMenteeListBasedOnSaasPolicy(userId, isAMentor) {
+	static async filterMenteeListBasedOnSaasPolicy(userId, isAMentor, organization_ids = []) {
 		try {
+			let extensionColumns = isAMentor ? await mentorQueries.getColumns() : await menteeQueries.getColumns()
+			// check for external_mentee_visibility else fetch external_mentor_visibility
+			extensionColumns = extensionColumns.includes('external_mentee_visibility')
+				? ['external_mentee_visibility', 'organization_id']
```
joffinjoy, line 1207: We haven't actually added these columns to the tables right? How is this code valid if that is the case?

adithyadinesh0412: Yes, @joffinjoy  We haven't added these fields as of now. 
But Akash suggested to do it this way so that once we add the fields it will work. 
So here also Im checking if we have field `external_mentee_visibility` . If the field is present, it will take data from there else it will go with `external_mentor_visibility `.

adithyadinesh0412: Yes, @joffinjoy  We haven't added these fields as of now. 
But Akash suggested to do it this way so that once we add the fields it will work. 
So here also Im checking if we have field `external_mentee_visibility` . If the field is present, it will take data from there else it will go with `external_mentor_visibility `.

PR #593, opened 2024-02-15
Katha #1168 - fix for question order in forms.

sumanvpacewisdom:


src/services/feedback.js
```diff
@@ -325,6 +325,15 @@ const getFeedbackQuestions = async function (formCode) {
 			let questions = await questionsQueries.find({
 				id: QuestionSet.questions,
 			})
+			const questionIndexMap = new Map()
+			QuestionSet.questions.forEach((id, index) => {
```
joffinjoy, line 329: This questionSet variable shouldn't start with a Captial letter. Make it camelCase.

PR #609, opened 2024-02-19
added meeting link for session creator

nevil-mathew:
fix for https://katha.shikshalokam.org/bug-view-1242.html

### DIFF
--- a/src/services/sessions.js
+++ b/src/services/sessions.js
@@ -861,7 +861,7 @@ module.exports = class SessionsHelper {
 				}
 			}
 
-			if (userId != sessionDetails.mentor_id) {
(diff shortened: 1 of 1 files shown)


src/services/sessions.js
```diff
@@ -861,7 +861,7 @@ module.exports = class SessionsHelper {
 				}
 			}
 
-			if (userId != sessionDetails.mentor_id) {
+			if (userId != sessionDetails.mentor_id && userId != sessionDetails.created_by) {
 				delete sessionDetails?.meeting_info?.link
```
rakeshSgr, line 865: @nevil-mathew Why we are deleting meeting_info ?

nevil-mathew: That's the existing flow, i just updated it to not delete those info if the creator is getting the details.

PR #637, opened 2024-02-23
filter list issue fixed

adithyadinesh0412:


src/services/mentees.js
```diff
@@ -925,8 +925,42 @@ module.exports = class MenteesHelper {
 						relatedOrgs = userOrgDetails.data.result.related_orgs
 					}
 					if (visibilityPolicy === common.ASSOCIATED) {
-						organization_ids.push(...relatedOrgs)
+						const associatedAdditionalFilter =
+							filter_type.toLowerCase() == common.MENTEE_ROLE
```
rakeshSgr, line 929: @adithyadinesh0412  filter_type.toLowerCase() at every condition is not good, set it at the beginning

src/services/mentees.js
```diff
@@ -925,8 +925,42 @@ module.exports = class MenteesHelper {
...
+											[Op.ne]: 'CURRENT',
+										},
+								  }
+
+						const organizationExtension = await organisationExtensionQueries.findAll(
+							{
+								[Op.and]: [
+									{
+										organization_id: {
+											[Op.in]: [...relatedOrgs, orgPolicies.organization_id],
```
rakeshSgr, line 946: @adithyadinesh0412 remove this from here orgPolicies.organization_id

add  this to  organization_ids []  by default

src/services/mentees.js
joffinjoy: @adithyadinesh0412 And rename this filter_type variable to filterType

PR #652, opened 2024-03-12
Issue fix for story - 1292

sumanvpacewisdom:
Mentor not able to start session after updating the session.


src/services/sessions.js
```diff
@@ -524,6 +524,11 @@ module.exports = class SessionsHelper {
 						await this.removeMentees(sessionId, menteesToRemove, bodyData.time_zone)
 					}
 				}
+
+				if (bodyData.start_date !== Number(sessionDetail.start_date)) {
```
rakeshSgr, line 528: @sumanvpacewisdom you should update the status only when, start_date > sessionDetail.start_date
no need of updating the status every time

PR #670, opened 2024-03-20
Fix for audit issue, Bug - 1154

sumanvpacewisdom:


src/services/platform.js
```diff
@@ -13,14 +13,18 @@ module.exports = class platformHelper {
 	 * @returns {Promise<Object>} - A promise that resolves with the application configuration.
 	 * @throws {Error} - Throws an error if there's an issue during configuration retrieval.
 	 */
-	static async getConfig() {
+	static async getConfig(decodedToken) {
 		try {
+			if (!decodedToken) {
```
rakeshSgr, line 18: @sumanvpacewisdom this not the right way to handle the tokenized API

src/database/migrations/20240321091346-platform-config-role-permission.js
```diff
@@ -0,0 +1,112 @@
...
+				},
+			})
+			await rowToDelete.destroy()
+		} catch (error) {
+			console.error('Error deleting row:', error)
+		}
+		try {
+			const rolePermissionsData = [
+				{
+					role_title: common.MENTEE_ROLE,
```
rakeshSgr, line 43: @sumanvpacewisdom we don't have ALL role?, if we introduce new role, again we need to migrations for that

PR #671, opened 2024-03-20
Fix for audit issue ,Bug - 1165

sumanvpacewisdom:


src/services/mentees.js
```diff
@@ -184,6 +184,11 @@ module.exports = class MenteesHelper {
 
 			let mySessions = await this.getMySessions(page, limit, search, userId)
 
+			mySessions.rows = mySessions.rows.map((session) => {
```
rakeshSgr, line 187: @sumanvpacewisdom remove it from the query

src/database/queries/sessions.js
```diff
@@ -593,10 +593,11 @@ exports.getUpcomingSessions = async (page, limit, search, userId) => {
 	}
 }
 
-exports.findAndCountAll = async (filter, options = {}) => {
+exports.findAndCountAll = async (filter, attributes, options = {}) => {
```
rakeshSgr, line 596: @sumanvpacewisdom set default attributes  ={}, so that it won't break, if we didn't pass anything

rakeshSgr: @sumanvpacewisdom add attributes as last argument otherwise it will break existing reference

PR #672, opened 2024-03-20
updated and improved validation logic

nevil-mathew:


src/validators/v1/issues.js
```diff
@@ -6,5 +6,8 @@ module.exports = {
 			.withMessage('description field is empty')
 			.isString()
 			.withMessage('description is invalid')
+			.not()
+			.matches(/(\b)(on\S+)(\s*)=|javascript:|<(|\/|[^\/>][^>]+|\/[^>][^>]+)>/gi)
```
rakeshSgr, line 10: @nevil-mathew what all things it will reject here

rakeshSgr: is a standerd regex ?

src/constants/blacklistConfig.js
```diff
@@ -0,0 +1,93 @@
...
+	update: ['id', 'organization_id', 'created_at', 'updated_at', 'created_by', 'updated_by'],
+}
+
+const profile = {
+	create: [
+		'user_id',
+		'rating',
+		'visibility',
+		'visible_to_organizations',
+		'external_session_visibility',
```
rakeshSgr, line 29: @nevil-mathew password we should not allow  to update, please check in user service also

src/validators/v1/issues.js
nevil-mathew: @rakeshSgr please refer this
https://www.regextester.com/110397

src/constants/blacklistConfig.js
nevil-mathew: mentoring service is not storing password, it should be rejected in user service

PR #678, opened 2024-03-28
added additional validation methods

nevil-mathew:


src/constants/common.js
```diff
@@ -131,4 +131,8 @@ module.exports = {
 	},
 	PUSH: 'PUSH',
 	POP: 'POP',
+	AUTH_METHOD: {
+		JWT_ONLY: 'jwt_only',
```
rakeshSgr, line 135: @nevil-mathew change to JWT: true,

src/middlewares/authenticator.js
```diff
@@ -97,7 +97,36 @@ module.exports = async function (req, res, next) {
...
+		if (process.env.AUTH_METHOD == common.AUTH_METHOD.USER_SERVICE) {
+			try {
+				const userBaseUrl = process.env.USER_SERVICE_HOST + process.env.USER_SERVICE_BASE_URL
+				const validateSessionEndpoint = userBaseUrl + endpoints.VALIDATE_SESSIONS
+				const reqBody = {
+					token: authHeader,
+				}
+				const isSessionActive = await requests.post(validateSessionEndpoint, reqBody, '', true)
 
+				if (isSessionActive.data.responseCode == 'UNAUTHORIZED') {
```
rakeshSgr, line 109: @nevil-mathew Is there any other response from the user service?

src/constants/common.js
nevil-mathew: @rakeshSgr The reason for not going with Boolean was to leave scope for future upgrades, like if we need to add more auth methods.
And the change to "JWT: JWT" wouldn't be fully correct because, even in the second method we need JWT, because the mentoring service won't work without the decoded token.

PR #688, opened 2024-04-03
added views init in app.js

nevil-mathew:


src/generics/materializedViews.js
```diff
@@ -357,11 +357,34 @@ const triggerPeriodicViewRefresh = async () => {
 		console.log(err)
 	}
 }
+const checkAndCreateMaterializedViews = async () => {
+	const allowFilteringEntityTypes = await getAllowFilteringEntityTypes()
+	const entityTypesGroupedByModel = await groupByModelNames(allowFilteringEntityTypes)
 
+	const query = 'select matviewname from pg_matviews;'
```
rakeshSgr, line 364: @nevil-mathew will the table name remain same?

PR #727, opened 2024-05-06
add: added client-cloud-services

nevil-mathew:


docker-compose-dev.yml
```diff
@@ -52,88 +40,111 @@ services:
...
       - elevate_net
+    deploy:
+      resources:
+        limits:
+          cpus: '4.0' # Set CPU limit to 4 cores
+          memory: '200M' # Set memory limit to 2 GB
   notification:
     build: '../notification/'
-    image: elevate/notification:1.0
+    image: elevate/notification:2.4
```
rakeshSgr, line 91: @nevil-mathew its refer to local image

src/.env.sample
```diff
@@ -182,3 +182,21 @@ MINIMUM_DURATION_FOR_AVAILABILITY=15
 
 #True to allow multiple bookings
 MULTIPLE_BOOKING=false
+
+
+# Cloud storage provider in azure, aws, gcloud, oci or s3
+CLOUD_STORAGE_PROVIDER=aws
+# Cloud storage identity [Azure Account Name, AWS Access Key, GCP Client Email, OCI S3 Access Key or S3 Access Key]
+CLOUD_STORAGE_ACCOUNTNAME=AKIAIOSFODNN7EXAMPLE
```
rakeshSgr, line 190: @nevil-mathew use dummy values

src/.env.sample
```diff
@@ -182,3 +182,21 @@ MINIMUM_DURATION_FOR_AVAILABILITY=15
 
 #True to allow multiple bookings
 MULTIPLE_BOOKING=false
+
+
+# Cloud storage provider in azure, aws, gcloud, oci or s3
+CLOUD_STORAGE_PROVIDER=aws
```
rakeshSgr, line 188: @nevil-mathew remove old env's
DEFAULT_AZURE_CONTAINER_NAME

docker-compose-dev.yml
nevilmathew: It's not the main docker compose file, it's untouched.
This docker compose is intended to be used for development.

nevil-mathew: This is not the main docker compose file, it is to be used for local development

src/.env.sample
nevil-mathew: These are dummy values

PR #746, opened 2024-05-23
Qa Bugs related issues - 1346 , 1350

sumanvpacewisdom:


src/locales/en.json
```diff
@@ -170,5 +170,6 @@
 	"CSV_UPLOADED_SUCCESSFULLY": "Sample Session Creation CSV Uploaded Successfully",
 	"CSV_UPDATE_FAILED": "Sample Session Creation CSV Upload Failed",
 	"SESSION_CSV_UPLOADED_FAILED": "Bulk Session Creation CSV Upload Failed",
-	"CSV_ROW_LIMIT_EXCEEDED": "CSV row limit exceeded. Please reduce the number of rows in your CSV file."
+	"CSV_ROW_LIMIT_EXCEEDED": "You can only upload 20 sessions at a time",
```
nevil-mathew, line 173: The limit is configurable right? So how can we hardcode this in the response?

nevil-mathew (changes requested):

nevil-mathew (changes requested):
Will this work if the mentee UID is email ?

src/services/sessionInvite.js
```diff
@@ -326,6 +328,13 @@ module.exports = class UserInviteHelper {
 				} else {
 					session.mentor_id = mentorId.result.id
 				}
+				if (
+					session.type == common.SESSION_TYPE.PRIVATE &&
+					!session.mentees.some((item) => typeof item === 'number')
```
nevil-mathew, line 333: Will this work if the mentee UID is email ?

sumanvpacewisdom: i am checking it after fetching the mentee id , so i have also set that if mentee exists replace with mentee id and if the mentee is invalid send the same string .

src/locales/en.json
sumanvpacewisdom: will change it

nevil-mathew (approved):

PR #779, opened 2024-06-20
Global Search FilterList Changes Based On Model Names 

sumanvpacewisdom:


src/services/mentees.js
```diff
@@ -902,7 +915,7 @@ module.exports = class MenteesHelper {
 			let organizationIds = []
 			filterType = filterType.toLowerCase()
 			const attributes =
-				filterType == common.MENTEE_ROLE
+				filterType == common.MENTEE_ROLE || common.SESSION
```
nevil-mathew, line 918: If its session, we should get the session policies right?

src/services/mentees.js
```diff
@@ -914,7 +927,7 @@ module.exports = class MenteesHelper {
 			)
 
 			const visibilityPolicy =
-				filterType == common.MENTEE_ROLE
+				filterType == common.MENTEE_ROLE || common.SESSION
```
nevil-mathew, line 930: If its session, we should get the session policies right?

src/services/mentees.js
sumanvpacewisdom: @nevil-mathew , external_mentee_visbility === external_session_visibility 
so based on that i did migrations also , so checking mentee is also fine it is same as session

nevil-mathew (approved):

PR #787, opened 2024-06-24
Changes related to sessions search filters

sumanvpacewisdom:


src/services/sessionInvite.js
```diff
@@ -461,13 +462,13 @@ module.exports = class UserInviteHelper {
 					}
 				}
 				const containsUserId = session.mentees.includes(userId)
-				if (!containsUserId && session.mentees.length >= 5) {
+				if (!containsUserId && session.mentees.length >= 6) {
```
rakeshSgr, line 465: @sumanvpacewisdom  use the existing config for this

PR #798, opened 2024-06-27
default-rule validators changes

sumanvpacewisdom:


src/validators/v1/default-rule.js
```diff
@@ -54,6 +51,7 @@ module.exports = {
 				.withMessage('Target field is required and must be a string')
 
 			req.checkBody('is_target_from_sessions_mentor')
+				.optional()
```
nevil-mathew, line 54: Remove checkBody for read, only id is needed

nevil-mathew: and id is optional

src/validators/v1/default-rule.js
```diff
@@ -21,7 +22,11 @@ module.exports = {
 
 			req.checkBody('field_configs').optional().isJSON().withMessage('Field configs must be a valid JSON')
 
-			req.checkBody('matching_operator').isString().notEmpty().withMessage('Matching operator must be a string')
+			req.checkBody('matching_operator')
+				.isString()
+				.notEmpty()
+				.isIn(['=', '!=', '>', '<', '>=', '<='])
```
nevil-mathew, line 28: Update this to words, eg: "equals"

src/validators/v1/default-rule.js
nevil-mathew, line 1: req.checkBody('role_config').isObject()

is .isObject() valid?

nevil-mathew: and add isIn to type as well, with allowed values in session, and mentor

nevil-mathew (changes requested):

sumanvpacewisdom:
@nevil-mathew Updated the commented changes , please review and let me know

src/validators/v1/default-rule.js
```diff
@@ -0,0 +1,129 @@
...
+					'equals',
+					'not equals',
+					'greater than',
+					'less than',
+					'greater than or equals',
+					'less than or equals',
+				])
+				.withMessage(
+					'Matching operator must be one of "equals", "not equals", "greater than", "less than", "greater than or equals", "less than or equals"'
+				)
```
nevil-mathew, line 42: It should match from these:
	['equals', '='],
	['notEquals', '!='],
	['contains', '@>'],
	['containedBy', '<@'],
	['overlap', '&&'],
	['lessThan', '<'],
	['greaterThan', '>'],
	['lessThanOrEqual', '<='],
	['greaterThanOrEqual', '>='],
so please remove the spaces in between.

nevil-mathew (approved):

PR #805, opened 2024-07-01
Bug fix changes - 1385,1387,1288,1400,1404

sumanvpacewisdom:


src/assets/BulkSessionCreationNew.csv
```diff
@@ -1,5 +1,6 @@
-Action,id,title,description,type,Mentor(Email/Mobile Num),Mentees(Email/Mobile Num),Date(DD-MM-YYYY),Time Zone(IST/UTC),Time (24 hrs),Duration(Min),recommended_for,categories,medium,Meeting Platform,Meeting Link or Meeting ID,Meeting Passcode (if needed)
+Action,id,title,description,type,Mentor(Email),Mentees(Email),Date(DD-MM-YYYY),Time Zone(IST/UTC),Time (24 hrs),Duration(Min),recommended_for,categories,medium,Meeting Platform,Meeting Link,Meeting Passcode (if needed)
```
rakeshSgr, line 1: @sumanvpacewisdom why mobile Num is removed

src/services/sessionInvite.js
```diff
@@ -213,7 +214,7 @@ module.exports = class UserInviteHelper {
 					},
 				}
 
-				parsedCSVData.push({
+				const parsedRow = {
```
rakeshSgr, line 217: @sumanvpacewisdom what is the change here ?

src/services/sessionInvite.js
```diff
@@ -229,7 +230,25 @@ module.exports = class UserInviteHelper {
 					categories: categoriesList,
 					medium: mediumList,
 					meeting_info: meetingInfo,
-				})
+				}
+
+				// Transform custom fields into an array format
+				const customEntities = {}
+				for (const [key, value] of Object.entries(customFields)) {
```
rakeshSgr, line 237: @sumanvpacewisdom make it generic function and add it in utils

src/services/sessionInvite.js
sumanvpacewisdom: @rakeshSgr Intially we were hardcoding the status csv with what ever we had , but now we can add entity types and for certain organization , so i have made it dynamically

src/services/sessionInvite.js
```diff
@@ -296,8 +323,16 @@ module.exports = class UserInviteHelper {
 						}
 					}
 					//Default Validation
-					const validateDefaultBBB = () => {
-						setMeetingInfo(common.MEETING_VALUES.BBB_LABEL, common.BBB_VALUE)
+					const validateDefaultBBB = async () => {
+						if (process.env.DEFAULT_MEETING_SERVICE === common.BBB_VALUE) {
+							setMeetingInfo('', '')
```
rakeshSgr, line 328: @sumanvpacewisdom its common statement for both condition keep it outside of if 
setMeetingInfo('', '')

src/assets/BulkSessionCreationNew.csv
sumanvpacewisdom: @rakeshSgr Products have changed the csv and asked to change , we are using only email for now is what they said

src/services/sessionInvite.js
```diff
@@ -296,8 +323,16 @@ module.exports = class UserInviteHelper {
 						}
 					}
 					//Default Validation
-					const validateDefaultBBB = () => {
-						setMeetingInfo(common.MEETING_VALUES.BBB_LABEL, common.BBB_VALUE)
+					const validateDefaultBBB = async () => {
+						if (process.env.DEFAULT_MEETING_SERVICE === common.BBB_VALUE) {
```
rakeshSgr, line 327: @sumanvpacewisdom you need to statusMessage for else only, change the logic

src/services/sessionInvite.js
```diff
@@ -659,29 +764,19 @@ module.exports = class UserInviteHelper {
...
-				meeting_info: item.meeting_info,
-			}))
+
+			const SessionBodyData = rowsWithStatus.map((item) => {
+				const { custom_entities, ...restOfItem } = item
+
+				// Add each key-value pair from custom_entities to the main session object
+				if (custom_entities) {
+					for (const [key, value] of Object.entries(custom_entities)) {
+						restOfItem[key] = value
```
rakeshSgr, line 774: @sumanvpacewisdom use better naming convention

src/services/sessionInvite.js
```diff
@@ -659,29 +764,19 @@ module.exports = class UserInviteHelper {
...
+					}
+				}
+
+				return restOfItem
```
rakeshSgr, line 778: @sumanvpacewisdom use proper naming convention every where

src/services/sessionInvite.js
```diff
@@ -738,54 +833,61 @@ module.exports = class UserInviteHelper {
...
+							mappedKey = 'Time (24 hrs)'
+							mappedRow[mappedKey] = restOfRow[key]
+							break
+						case 'duration':
+							mappedKey = 'Duration(Min)'
+							mappedRow[mappedKey] = restOfRow[key]
+							break
+						default:
+							mappedRow[key] = restOfRow[key] // Use the original key
+					}
```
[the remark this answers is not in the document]
sumanvpacewisdom: @rakeshSgr , It will be mapped based on existing name , because entity_types will be given how it is there in the db and it won't accept if the entoties are not according to the entity labels

PR #810, opened 2024-07-01
Entity search

rakeshSgr:


src/controllers/v1/entity.js
```diff
@@ -81,4 +81,22 @@ module.exports = class Entity {
...
+
+	/**
+	 * entity details
+	 * @method
+	 * @name details
+	 * @param {Object} req - request data.
+	 * @returns {JSON} - entities.
+	 */
+
+	async details(req) {
```
aks30, line 93: @rakeshSgr Let's rename it to list

src/controllers/v1/entity.js
```diff
@@ -81,4 +81,22 @@ module.exports = class Entity {
...
+		try {
+			req.decodedToken = {}
+			req.decodedToken.id = 0
```
aks30, line 96: Not sure what is this about?

src/database/queries/entity.js
```diff
@@ -55,4 +57,21 @@ module.exports = class UserEntityData {
 			return error
 		}
 	}
+
+	static async getAllEntities(filters, attributes, page, limit, search) {
```
aks30, line 61: Can we extend and reuse the function - findAllEntities

src/services/entity.js
```diff
@@ -204,4 +204,65 @@ module.exports = class EntityHelper {
...
+	 * @param {String} searchText - search label in entity.
+	 * @param {Integer} page -  page no.
+	 * @param {Integer} pageSize -  page limit per api.
+	 * @returns {JSON} - Entity search matched response.
+	 */
+	static async details(query, userId, searchText, pageNo, pageSize) {
+		try {
+			let entityType = query.entity_type_id ? query.entity_type_id : ''
+			let filter
+			if (query.read_user_entity == true) {
```
aks30, line 223: What does the key read_user_entity mean? When will it be true and false?

aks30 (changes requested):
Reviewed on 1 July

src/services/entity.js
rakeshSgr: read_user_entity = = true, it reads requested user created entity and entity created by id 0, 
which means default entities,
The same logic is there in existing read api also

src/database/queries/entity.js
rakeshSgr: that returns only data,  we need data with count as well.  we can do that

rakeshSgr: we have pagination so it won't return all data,  we can't use it

src/controllers/v1/entity.js
rakeshSgr: done

src/controllers/v1/entity.js
rakeshSgr: resolved

src/services/entity.js
```diff
@@ -204,4 +204,65 @@ module.exports = class EntityHelper {
...
+				filter = {
+					[Op.or]: [
+						{
+							created_by: '0',
+						},
+						{
+							created_by: userId,
```
aks30, line 230: Will there be ever a case when regular user will have put an entry in entity table that we have to filter by his userId?

aks30 (changes requested):
Reviewed on 2 July

src/services/entity.js
rakeshSgr: currently its not require,  in the beginning  user can also create entities in that this filter was added

rakeshSgr: frontend not using this

src/controllers/v1/entity.js
```diff
@@ -81,4 +81,20 @@ module.exports = class Entity {
...
+	 * entity list
+	 * @method
+	 * @name list
+	 * @param {Object} req - request data.
+	 * @returns {JSON} - entities.
+	 */
+
+	async list(req) {
+		try {
+			return await entityService.list(req.query, req.decodedToken.id, req.searchText, req.pageNo, req.pageSize)
```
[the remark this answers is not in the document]
rakeshSgr: removed this, not require

PR #816, opened 2024-07-02
Fix: Maintain order while iterating over extensionDetails

nevil-mathew:


src/services/mentors.js
```diff
@@ -944,10 +945,7 @@ module.exports = class MentorsHelper {
 			return responses.successResponse({
 				statusCode: httpStatusCode.ok,
 				message: userDetails.message,
-				result: {
```
rakeshSgr, line 947: @nevil-mathew why response is changed

nevil-mathew: The response structure is same for 'extensionDetails', so removed the unnecessary mapping.

rakeshSgr: @nevil-mathew where you are getting  extensionDetails.count

PR #820, opened 2024-07-03
updated: validations and api-doc

nevil-mathew:


src/services/default-rule.js
```diff
@@ -42,61 +52,58 @@ module.exports = class DefaultRuleHelper {
...
-			})
+			errors.push({ param: 'requester_field', msg: 'Invalid requester_field' })
 		}
 
-		if (errors.length !== 0) {
-			return {
-				isValid: false,
-				errors,
+		if (validTargetField.length > 0 && validRequesterField.length > 0) {
+			if (validTargetField[0]?.data_type !== validRequesterField[0]?.data_type) {
```
rakeshSgr, line 73: @nevil-mathew if validTargetField has  more than one field, then won't check for those ?

src/services/default-rule.js
```diff
@@ -42,61 +52,58 @@ module.exports = class DefaultRuleHelper {
...
+				errors.push({
+					param: 'target_field,requester_field',
+					msg: 'Data types of target_field and requester_field should match',
+				})
+			} else {
+				const operatorValidation = {
```
rakeshSgr, line 79: @nevil-mathew checking for first field is enough?

nevil-mathew: We're checking if the target_field and requester_field have the same data_type before this, when we reach this condition both will have same data_type

src/services/default-rule.js
nevil-mathew: We're only checking other conditions if the target and requester fields are valid.
Else we don't need to check the data type, operator validity.

PR #824, opened 2024-07-04
Entity type validations

nevil-mathew:


src/database/queries/mentorExtension.js
```diff
@@ -182,12 +172,12 @@ module.exports = class MentorExtensionQueries {
 				additionalFilter = `${searchFilter.whereClause}`
 			}
 
-			let filterClause = filterConditions.length > 0 ? ` ${filterConditions.join(' AND ')}` : ''
+			const filterClause = filter.query.length > 0 ? `${filter.query}` : ''
```
rakeshSgr, line 175: @nevil-mathew check filter.query exist or not then check the length

nevil-mathew: Added safe chaining, but this response is returned from internal function, so this will always have `filter.query`.

src/validators/v1/entity-type.js
```diff
@@ -88,6 +131,40 @@ module.exports = {
...
+		req.checkBody('required').optional().isBoolean().withMessage('required is invalid, must be boolean')
+
+		req.checkBody('regex')
+			.optional()
+			.notEmpty()
+			.withMessage('regex field is empty')
+			.isString()
+			.withMessage('regex is invalid,must be string')
+
+		req.checkBody('status')
```
rakeshSgr, line 146: @nevil-mathew it won't break any existing use case right?

PR #830, opened 2024-07-11
Adding organization field to filterList API

sumanvpacewisdom:


src/services/mentees.js
```diff
@@ -958,6 +958,10 @@ module.exports = class MenteesHelper {
 				}
 			}
 
+			if (organization.toLowerCase() === common.FALSE) {
```
rakeshSgr, line 961: @sumanvpacewisdom first why ur fetching the org details if its not require in case of organisation is set false

PR #840, opened 2024-07-18
Bug Fix : 1553 , 1554 Adding LowerCase Validation

sumanvpacewisdom:


src/services/entity-type.js
```diff
@@ -58,6 +59,7 @@ module.exports = class EntityHelper {
 
 	static async update(bodyData, id, loggedInUserId, orgId, roles) {
 		bodyData.updated_by = loggedInUserId
+		bodyData.value = bodyData.value.toLowerCase()
```
nevil-mathew, line 62: Can you check if the value is mandatory for the update?

nevil-mathew:
Approved

src/services/default-rule.js
```diff
@@ -47,13 +47,15 @@ module.exports = class DefaultRuleHelper {
 				value: bodyData.target_field,
 				model_names: { [Op.contains]: [modelName] },
 				required: true,
+				allow_filtering: true,
```
rakeshSgr, line 50: @sumanvpacewisdom we are only validating the allow_filtering:true, for other entity types do we have validation ?

PR #845, opened 2024-07-22
Fix : Deleting_Status_Entity

sumanvpacewisdom:


src/database/migrations/20240722123546-deleting_status_entity.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+	async up(queryInterface, Sequelize) {
+		// Delete entities with label 'Unfulfilled' and value 'UNFULFILLED'
+		await queryInterface.bulkDelete('entities', { label: 'Unfulfilled', value: 'UNFULFILLED' }, {})
+	},
+
+	async down(queryInterface, Sequelize) {
```
aks30, line 10: @sumanvpacewisdom How is this valid down operation. Don't we need to add entity type id in entity table in case we have to insert?

PR #847, opened 2024-07-23
Fix : Skip Default Rule Validation for mentees

sumanvpacewisdom:


src/services/sessions.js
```diff
@@ -903,30 +903,36 @@ module.exports = class SessionsHelper {
...
-					message: 'PROFILE_NOT_UPDATED',
-					statusCode: httpStatusCode.bad_request,
-					responseCode: 'CLIENT_ERROR',
-				})
-			}
+			let sessionAttendee = await sessionAttendeesQueries.findOne({
+				session_id: sessionDetails.id,
+				mentee_id: userId,
+			})
+			if (!sessionAttendee) {
```
nevil-mathew, line 910: Why is this added as an outer clause?

nevil-mathew (changes requested):

nevil-mathew (approved):

PR #848, opened 2024-07-23
added: support for ids in string.

nevil-mathew:


src/database/migrations/20240716111210-update-user-id-to-string.js
```diff
@@ -0,0 +1,85 @@
...
+			})
+		} catch (error) {
+			console.log(error)
+			throw error
+		}
+	},
+
+	down: async (queryInterface, Sequelize) => {
+		// Revert changes back to INTEGER
+		await queryInterface.changeColumn('availabilities', 'user_id', {
```
rakeshSgr, line 55: @nevil-mathew set allowNull: false

docker-compose-dev.yml
```diff
@@ -183,7 +183,7 @@ services:
   #     - elevate_net
 
   citus:
-    image: citusdata/citus:11.2.0
+    image: citusdata/citus:12.1
```
rakeshSgr, line 186: @nevil-mathew why this is changed?

src/database/migrations/20240716111210-update-user-id-to-string.js
```diff
@@ -0,0 +1,85 @@
+'use strict'
+
+module.exports = {
+	up: async (queryInterface, Sequelize) => {
+		try {
+			// Change columns to STRING
+			await queryInterface.changeColumn('availabilities', 'user_id', {
+				type: Sequelize.STRING,
```
rakeshSgr, line 8: @nevil-mathew check is there any test case we have ?  if its present, do we need to change in mock data in that case
@nevil-mathew API doc request and response check if there any change required

docker-compose-dev.yml
nevil-mathew: To keep parity with DEV/QA server.

src/database/migrations/20240716111210-update-user-id-to-string.js
nevil-mathew: updated

PR #851, opened 2024-07-24
Fix:DefaultRule validation for upcomingSession API

sumanvpacewisdom:


src/controllers/v1/mentors.js
```diff
@@ -30,7 +30,9 @@ module.exports = class Mentors {
 				req.searchText,
 				req.params.menteeId ? req.params.menteeId : req?.decodedToken?.id,
 				req.query,
-				isAMentor(req.decodedToken.roles)
+				isAMentor(req.decodedToken.roles),
```
rakeshSgr, line 33: @sumanvpacewisdom is this api will be calling only by mentor ?

PR #853, opened 2024-07-26
Fix : 1558 fetching profile table instead of views

sumanvpacewisdom:


src/helpers/defaultRules.js
```diff
@@ -100,7 +100,7 @@ exports.defaultRulesFilter = async function defaultRulesFilter({
 }) {
 	try {
 		const [userDetails, defaultRules] = await Promise.all([
-			getUserDetailsFromView(requesterId, isAMentor(roles)),
+			getUserDetails(requesterId, isAMentor(roles)),
```
rakeshSgr, line 103: @sumanvpacewisdom one place only ?

sumanvpacewisdom: @rakeshSgr Among two defaultRules functions only one is using the view , rest all are using user table directly

PR #854, opened 2024-07-26
Fix : updateOrganization API adding bad request

sumanvpacewisdom:


src/validators/v1/organization.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+	update: (req) => {
+		req.checkBody('mentee_feedback_question_set').notEmpty().withMessage('mentee_feedback_question_set is empty')
```
rakeshSgr, line 3: @sumanvpacewisdom It should be optional, if either one is passed, both are required 

otherwise how do you update the other fields  ?

src/validators/v1/organization.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+	update: (req) => {
+		req.checkBody('mentee_feedback_question_set')
+			.notEmpty()
+			.withMessage('mentee_feedback_question_set is required')
+			.matches(/^[a-zA-Z0-9_]+$/)
```
rakeshSgr, line 6: @sumanvpacewisdom it should be same validation, what we allow while creating question sets

sumanvpacewisdom: @rakeshSgr Its same , i have verified and added this

PR #869, opened 2024-07-31
FIx:1636 Adding Validation for DefaultRule columns

sumanvpacewisdom:


src/validators/v1/default-rule.js
```diff
@@ -58,6 +58,10 @@ module.exports = {
 			.withMessage('The requester_roles field must have at least one role.')
 			.custom((roles) => roles.every((role) => typeof role === 'string'))
 			.withMessage('All elements in the requester_roles array must be strings.')
+			.custom((roles) => roles.every((role) => /^[a-z_]+$/.test(role)))
```
rakeshSgr, line 61: @sumanvpacewisdom it allows only small letter, allow Caps also

sumanvpacewisdom: roles are created with only small letters, based on that i have put the validations @rakeshSgr

PR #872, opened 2024-08-02
Fix : Validations and API Doc Updation

sumanvpacewisdom:


src/validators/v1/question-set.js
```diff
@@ -8,7 +8,11 @@ const { filterRequestBody } = require('../common')
 const { questionSet } = require('@constants/blacklistConfig')
 module.exports = {
 	read: (req) => {
-		req.checkParams('id').notEmpty().withMessage('id param is empty')
+		req.checkParams('id')
```
rakeshSgr, line 11: @sumanvpacewisdom we are moving from integer to string check, how the validation is done

sumanvpacewisdom: @rakeshSgr This is related to questionSet id , but we are converting user_id to string right

PR #894, opened 2024-08-08
merging mentor table with user table

nevil-mathew:


src/database/migrations/20240801074543-merge-mentor-user-extensions.js
```diff
@@ -0,0 +1,155 @@
...
+					allowNull: false,
+					defaultValue: false,
+				},
+				{ transaction }
+			)
+
+			// Copy data from mentor_extensions to user_extensions
+			await queryInterface.sequelize.query(
+				`
+        INSERT INTO user_extensions (
```
rakeshSgr, line 27: @nevil-mathew the user is mentee then he become mentor, in this case this statement will create duplicate entry in user_extensions ?

src/database/migrations/20240801074543-merge-mentor-user-extensions.js
```diff
@@ -0,0 +1,155 @@
...
+				{ transaction }
+			)
+
+			// Drop the mentor_extensions table
+			await queryInterface.dropTable('mentor_extensions', { transaction })
+		})
+		await materializedViewsService.checkAndCreateMaterializedViews()
+	},
+
+	down: async (queryInterface, Sequelize) => {
```
rakeshSgr, line 51: @nevil-mathew mentor_extensions view creation is missing

nevil-mathew: `await materializedViewsService.checkAndCreateMaterializedViews()`

This function will create all the necessary view. Also, we don't need mentor view since it's table is removed.

rakeshSgr: in down we don't have it, so can you add it

nevil-mathew: added

PR #899, opened 2024-08-08
converting user_id to string & validation

sumanvpacewisdom:


src/services/sessions.js
```diff
@@ -2216,6 +2216,7 @@ module.exports = class SessionsHelper {
 	 */
 	static async bulkUpdateMentorNames(mentorsId, mentorsName) {
 		try {
+			mentorsId = mentorsId.toString()
```
joffinjoy, line 2219: @param {Array} mentorsId

This states that mentorsId is an array, but you have applied toString(). Seems like a mismatch. Please clarify this. 

And if this is an array of mentor ids, rename this to mentorIds instead mentorsId.

joffinjoy (changes requested):
comments added.

PR #924, opened 2024-08-27
Updated: init migration to handle string org IDs

nevil-mathew:


src/configs/postgres.js
```diff
@@ -26,16 +26,16 @@ module.exports = {
 			deletedAt: 'deleted_at',
 			logging: false,
 		},
-		defaultOrgId: parseInt(defaultOrgId), // Convert to integer
+		defaultOrgId: defaultOrgId, // Convert to integer
```
rakeshSgr, line 29: @nevil-mathew change comments

PR #937, opened 2024-09-09
Deleting EntityTypes and Entities for KB

sumanvpacewisdom:


src/database/queries/entityType.js
```diff
@@ -159,4 +159,48 @@ module.exports = class UserEntityData {
...
+
+				// Step 4: Delete the entities and entityTypes
+				await Entity.destroy({
+					where: { entity_type_id: entityTypeIds },
+				})
+
+				await EntityType.destroy({
+					where: { id: entityTypeIds },
+				})
+				return result
```
nevil-mathew, line 197: Why are we manually deleting the entity types? Can you refer to the existing flow, where we are using hooks to delete the entities?
Can you check if it's possible here?

nevil-mathew: Instead of 
```
				await Entity.destroy({
					where: { entity_type_id: entityTypeIds },
				})
```
can we make use of hooks here?

nevil-mathew:
Approved

PR #950, opened 2024-09-18
Removing regex validation for label

sumanvpacewisdom:
Due to KB entity creations

### DIFF
--- a/src/validators/v1/entity.js
+++ b/src/validators/v1/entity.js
@@ -17,12 +17,7 @@ module.exports = {
 			.matches(/^[A-Za-z0-9_]*$/)
 			.withMessage('value is invalid, must not contain spaces')
 
-		req.checkBody('label')
-			.trim()
-			.notEmpty()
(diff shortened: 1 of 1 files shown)


src/validators/v1/entity.js
```diff
@@ -17,12 +17,7 @@ module.exports = {
 			.matches(/^[A-Za-z0-9_]*$/)
 			.withMessage('value is invalid, must not contain spaces')
 
-		req.checkBody('label')
-			.trim()
-			.notEmpty()
-			.withMessage('label field is empty')
-			.matches(/^[A-Za-z0-9 ()-]*$/)
-			.withMessage('label is invalid')
+		req.checkBody('label').trim().notEmpty().withMessage('label field is empty').withMessage('label is invalid')
```
rakeshSgr, line 20: @sumanvpacewisdom remove trim here

rakeshSgr: we should store exactly

PR #967, opened 2024-10-15
added: user info to the pending and getinfo api

nevil-mathew:


src/services/connections.js
```diff
@@ -82,6 +88,31 @@ module.exports = class ConnectionHelper {
 				})
 			}
 
+			const defaultOrgId = await getDefaultOrgId()
```
aks30, line 91: @nevil-mathew we can skip making an API call here if we just need default org id

aks30 (changes requested):
Reviewed on 15 Oct

src/services/connections.js
nevil-mathew: Updated `getDefaultOrgId()` to return from the environment if it exists.

src/services/connections.js
```diff
@@ -96,10 +127,47 @@ module.exports = class ConnectionHelper {
...
+				})
+			}
+			const friendIds = connections.rows.map((connection) => connection.friend_id)
+			let friendDetails = await menteeQueries.getUsersByUserIds(friendIds, {
+				attributes: ['user_id', 'name', 'designation', 'organization_id'],
+			})
+			const userExtensionsModelName = await menteeQueries.getModelName()
+
+			const uniqueOrgIds = [...new Set(friendDetails.map((obj) => obj.organization_id))]
+			friendDetails = await entityTypeService.processEntityTypesToAddValueLabels(
```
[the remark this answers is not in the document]
nevil-mathew: FE need the role to be shown in the message request tab.

PR #990, opened 2024-11-07
Doumentation Changes

sumanvpacewisdom:


documentation/3.1.0/native/envs/interface_env
```diff
@@ -1,16 +1,15 @@
 API_DOC_URL=/elevate-interface/api-doc
 APPLICATION_ENV=development
 APPLICATION_PORT=3569
-INSTALLED_PACKAGES=elevate-user elevate-mentoring elevate-scheduler
+INSTALLED_PACKAGES=elevate-user elevate-mentoring elevate-scheduler elevate-project elevate-entity-management elevate-self-creation-portal elevate-survey
```
nevil-mathew, line 4: Can you check the current env in dev or qa, to verify if we need to use other service packages.

sumanvpacewisdom: i have checked this env's work for 3.1.0 release code , its correct only @nevil-mathew

PR #1029, opened 2024-12-02
env_optimizations

sumanvpacewisdom:


src/envVariables.js
```diff
@@ -112,52 +135,79 @@ let enviromentVariables = {
...
 		message: 'Required session edit window timeout',
-		optional: false,
+		optional: true,
+		default: 0,
 	},
 	SESSION_MENTEE_LIMIT: {
 		message: 'Required session mentee limit',
-		optional: false,
+		optional: true,
+		default: 7,
```
nevil-mathew, line 171: Make this as 20

src/envVariables.js
```diff
@@ -287,11 +343,13 @@ let enviromentVariables = {
...
-		optional: false,
+		optional: true,
+		default: '9bbf4698e081176b4e38c3f04f9c0cff6ae80b1960e5ac868cc234ac8537aa66',
 	},
 	EMAIL_ID_ENCRYPTION_IV: {
 		message: 'Required Email ID Encryption IV',
-		optional: false,
+		optional: true,
+		default: 'a390328e496f025249a88384f318a536',
 	},
```
nevil-mathew, line 353: Please mark it as optional false and remove default values.

nevil-mathew (changes requested):

src/envVariables.js
```diff
@@ -55,31 +57,49 @@ let enviromentVariables = {
...
+		optional: process.env.DEFAULT_MEETING_SERVICE === 'OFF' ? true : false,
 	},
 	MEETING_END_CALLBACK_EVENTS: {
 		message: 'Required meeting end callback events',
 		optional: false,
 	},
 	BIG_BLUE_BUTTON_SECRET_KEY: {
 		message: 'Required big blue button secret key',
-		optional: false,
+		optional: process.env.DEFAULT_MEETING_SERVICE === 'OFF' ? true : false,
```
nevil-mathew, line 102: Can we update the BBB related keys optional check to
process.env.DEFAULT_MEETING_SERVICE === 'BBB' ? false : true

src/envVariables.js
```diff
@@ -112,52 +135,79 @@ let enviromentVariables = {
...
+		optional: true,
+		default: 'silly',
 	},
 	DISABLE_LOG: {
 		message: 'Required disable log level',
-		optional: false,
+		optional: true,
+		default: false,
 	},
 	DEFAULT_MEETING_SERVICE: {
```
nevil-mathew, line 159: Can we make this as optional and have default value as 'OFF'

sumanvpacewisdom: no we cannot.

nevil-mathew: Why not?

sumanvpacewisdom: here i gave default as OFF and it did not take in the next upcomming conditions which i have used 
so i need to mention int he env file so other conditions work.

PR #1035, opened 2024-12-03
added: image and user profile update integration with communication service

nevil-mathew:


src/helpers/communications.js
```diff
@@ -105,7 +175,11 @@ exports.createChatRoom = async (recipientUserId, initiatorUserId, initialMessage
 			if (!user.meta || !user.meta.communications_user_id) {
 				// Decrypt email and create user in communication service if `communications_user_id` is missing
 				user.email = await emailEncryption.decrypt(user.email)
-				await this.create(user.user_id, user.name, user.email, 'https://picsum.photos/200/200')
+				let userImage
+				if (user?.image) {
+					userImage = await utils.getDownloadableUrl(user.image)
```
rakeshSgr, line 180: @nevil-mathew image should download from user service right ?

src/services/connections.js
```diff
@@ -235,9 +235,8 @@ module.exports = class ConnectionHelper {
 				}
 			})
 
-			//To be removed later
-			connectionsWithDetails.forEach((detail) => {
-				detail.user_details.image = 'https://picsum.photos/200'
+			connectionsWithDetails.forEach(async (detail) => {
+				detail.user_details.image = await utils.getDownloadableUrl(detail.user_details.image)
```
rakeshSgr, line 239: @nevil-mathew download the image from user service

src/controllers/v1/profile.js
```diff
@@ -162,6 +162,33 @@ module.exports = class Mentees {
...
+	 * @param {Object} req.params - Route parameters.
+	 * @param {String} req.params.id - The mentor's ID.
+	 * @param {Object} req.decodedToken - Decoded token from authentication.
+	 * @param {String} req.decodedToken.id - The user's ID.
+	 * @param {String} req.decodedToken.organization_id - The user's organization ID.
+	 * @param {Array} req.decodedToken.roles - The user's roles.
+	 * @param {Boolean} isAMentor - Indicates whether the user is a mentor.
+	 * @returns {Promise<Object>} - The mentor's profile details.
+	 */
+	async details(req) {
```
rakeshSgr, line 179: @nevil-mathew why this API needed ?

nevil-mathew: Currently, we only have API to get mentor details, this API allows getting any user details. Will update the JS Doc to mention that.

src/helpers/communications.js
nevil-mathew: We have added images to the user extension table right, so we will use that image path to get a downloadable URL, which will be used in communication service to download the image and add it to rocket chat.

src/services/connections.js
nevil-mathew: We're using the newly added image in user extension to get the image URL.

PR #1050, opened 2024-12-13
replacing BIB_BLUE_BUTTON_BASE_URL to BIG_BLUE_BUTTON_BASE_URL

sumanvpacewisdom:


src/envVariables.js
```diff
@@ -137,7 +137,7 @@ let enviromentVariables = {
 		optional: process.env.ENABLE_EMAIL_FOR_REPORT_ISSUE === 'true' ? false : true,
 		default: 'user_issue_reported',
 	},
-	BIG_BLUE_BUTTON_BASE_URL: {
+	BIB_BLUE_BUTTON_BASE_URL: {
```
rakeshSgr, line 140: @suman Can you update the code BIG_BLUE_BUTTON_BASE_URL, so it will be proper

sumanvpacewisdom: then no need of merging this pr , its already having the same @rakeshSgr

rakeshSgr: but in code we still refer this BIB_BLUE_BUTTON_BASE_URL

rakeshSgr: update the code where ever we use this key

sumanvpacewisdom: @rakeshSgr ok

src/api-doc/mentoring-2.5-setup.md
```diff
@@ -366,7 +366,7 @@ USER_SERVICE_BASE_URL='/user/'
 BIG_BLUE_BUTTON_URL=https://dev.some.temp.org
 
 # Big blue button base URL
-BIB_BLUE_BUTTON_BASE_URL=/bigbluebutton/
+BIG_BLUE_BUTTON_BASE_URL=/bigbluebutton/
```
rakeshSgr, line 369: @sumanvpacewisdom for older version of docs, we should't make any env changes, the docker images dependent on those env keys only

PR #1063, opened 2024-12-30
milestone 3 development

sumanvpacewisdom:


src/constants/common.js
```diff
@@ -144,6 +144,7 @@ module.exports = {
 	BACK_OFF_RETRY_QUEUE: 600000,
 	tempFolderForBulkUpload: 'public/invites',
 	sessionOutputFile: 'output-session-creation',
+	reportOutputFile: 'output-report-data',
```
rakeshSgr, line 147: @sumanvpacewisdom will we have same file alway ?

PR #1069, opened 2024-12-31
enhancement and queries changes milestone

sumanvpacewisdom:


src/services/reports.js
```diff
@@ -133,67 +133,81 @@ module.exports = class ReportsHelper {
...
 				config: reportConfig.dataValues.config,
 			}
 
-			// Fetch report query
-			const reportQuery = await reportQueryQueries.findReportQueryByCode(reportCode)
-			let query = reportQuery.query
-
-			// Replace placeholders
+			// Prepare query replacements
+			const defaultLimit = common.pagination.DEFAULT_LIMIT
```
nevil-mathew, line 167: Do we need `defaultLimit` ? Don't we get the default (100) as the limit?

src/services/reports.js
```diff
@@ -204,8 +218,12 @@ module.exports = class ReportsHelper {
...
 			})
 		} catch (error) {
-			console.error('Error in getReportData:', error)
-			throw error
+			return responses.failureResponse({
+				message: 'INTERNAL_SERVER_ERROR',
+				statusCode: httpStatusCode.internal_server_error,
+				responseCode: 'SERVER_ERROR',
+				error: error.message,
+			})
```
nevil-mathew, line 226: In [generateAndUploadCSV](https://github.com/ELEVATE-Project/mentoring/blob/49135d03c78918057615556e4f9695502275ae78/src/services/reports.js#L233), the temporary file created during CSV generation is not being removed, which could lead to unnecessary file accumulation. Can we add logic to ensure the file is deleted after the upload is successful, and also handle clean-up in case the function encounters an error?

src/services/reports.js
```diff
@@ -133,67 +133,81 @@ module.exports = class ReportsHelper {
 		sortColumn,
 		sortType,
 		searchColumn,
-		searchValue
+		searchValue,
+		downloadCsv
 	) {
```
nevil-mathew, line 138: Add proper JS Doc for all functions

nevil-mathew (changes requested):

src/services/reports.js
sumanvpacewisdom: No @nevil-mathew

PR #1228, opened 2025-06-15
bug 2934, 2941

sumanvpacewisdom:


src/database/migrations/20250211135450-request_sessions_table.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
 				allowNull: false,
 			},
 			agenda: {
-				type: Sequelize.STRING,
+				type: Sequelize.STRING(2000),
```
rakeshSgr, line 33: @sumanvpacewisdom use TEXT field and validate in validators ( for agend 300 characters )

sumanvpacewisdom: Done

PR #1328, opened 2025-08-19
kafka-topic-added

MallanagoudaB:


src/envVariables.js
```diff
@@ -558,6 +558,11 @@ let enviromentVariables = {
 		optional: true,
 		default: 'mentor_has_changed',
 	},
+	KAFKA_HEALTH_CHECK_TOPIC: {
+		message: 'Required KAFKA_HEALTH_CHECK_TOPIC',
+		optional: false,
```
rakeshSgr, line 563: @MallanagoudaB if optional is false then don't set the default value

PR #1372, opened 2025-09-29
adding exipre request session api for internal access

rakeshSgr:


src/constants/common.js
```diff
@@ -56,6 +56,7 @@ module.exports = {
 		'/sessions/removeAllSessions',
 		'/mentoring/v1/users/add',
 		'/mentoring/v1/users/delete',
+		'/requestSessions/expire',
```
aks30, line 59: @rakeshSgr where is the endpoint called from what workflow?

rakeshSgr: When session requested created, creating a job in scheduler, so that once session requested is expired, it should call the this API to set that status as exipred

PR #1490, opened 2025-12-05
Enroll ap ifix

rakeshSgr:
### Key Changes

- **Cache Invalidation for Mentor Transitions**: Added cache cleanup when mentors transition to mentees, with defensive error handling for failed cache deletions
- **API Refactoring**: Renamed `_clearUserCacheForSessionCountChange()` to `_clearUserCache()` with simplified parameters (removed action parameter) and updated all call sites accordingly
- **Cache-First Architecture**: Expanded cache-first retrieval patterns across mentor/mentee/session/resource lookups; replaced direct database queries with `cacheHelper.getCacheOnly()` and `getCacheThenFallback()` patterns throughout session lifecycle
- **Resilient Cache Operations**: Wrapped all cache invalidations and lookups in try/catch blocks to ensure business logic continues gracefully on cache failures
- **Resource URL Handling**: Introduced `getResourceAccessibleUrl()` public method to normalize and standardize resource links into accessible URLs during session creation, details, and completion
- **Session Lifecycle Optimization**: Updated session operations (enroll, unenroll, update, delete, start, share) to perform intelligent cache operations with automatic database fallbacks
- **Enrollment Data Access Simplification**: Simplified `getEnrolledMentees()` signature by removing unused `userID` parameter, reducing API surface complexity
- **Attendee Unenrollment Enhancement**: Modified `unEnrollAllAttendeesOfSessions()` to return structured object containing both `deletedCount` and `deletedRecords`, enabling granular per-attendee cache invalidation
- **Per-Record Cache Cleanup**: Added targeted mentee cache invalidation during session unenrollment, iterating through deleted records to clear individual mentee cache entries
- **Data Sourcing Optimization**: Mentor and mentee metadata (names, emails) now preferentially sourced from cache with seamless database fallbacks for resilience

### Contribution Summary

| Rocky (rakeshSgr) | 10,055 | 0 |

### DIFF
--- a/src/requests/user.js
+++ b/src/requests/user.js
@@ -950,7 +950,7 @@ const getUserDetailedListUsingCache = async function (usersMap, tenantCode, dele
 				usersInfo.missingUserIds,
 				options,
 				tenantCode,
-				unscopped
+				(unscopped = true)
 			)
 			userDetails.push(...usersFromDb)
 		}
--- a/src/requests/user.js
+++ b/src/requests/user.js
@@ -950,7 +950,7 @@ const getUserDetailedListUsingCache = async function (usersMap, tenantCode, dele
 				usersInfo.missingUserIds,
 				options,
 				tenantCode,
-				(unscopped = true)
+				true
 			)
 			userDetails.push(...usersFromDb)
 		}
--- a/src/requests/user.js
+++ b/src/requests/user.js
@@ -950,7 +950,7 @@ const getUserDetailedListUsingCache = async function (usersMap, tenantCode, dele
 				usersInfo.missingUserIds,
 				options,
 				tenantCode,
-				true
+				unscopped
 			)
 			userDetails.push(...usersFromDb)
 		}
--- a/src/services/sessions.js
+++ b/src/services/sessions.js
@@ -1674,7 +1674,7 @@ module.exports = class SessionsHelper {
 			}
 
 			const canRetrieveMenteeList = userId == sessionDetails.created_by || userId == sessionDetails.mentor_id
-			sessionDetails.mentees = await getEnrolledMentees(sessionDetails.id, {}, userId, tenantCode)
+			sessionDetails.mentees = await getEnrolledMentees(sessionDetails.id, {}, tenantCode)
 
 			let sessionAccessorDetails
 			if (isInvited || sessionDetails.is_assigned || !mentorExtension) {
@@ -2404,7 +2404,7 @@ module.exports = class SessionsHelper {
 			}
 
 			// Clear user cache since sessions_attended count changed
-			await this._clearUserCache(userTokenData.user_id, tenantCode, orgCode)
+			await this._clearUserCache(userId, tenantCode, orgCode)
 
 			return responses.successResponse({
 				statusCode: httpStatusCode.accepted,
@@ -2488,6 +2488,11 @@ module.exports = class SessionsHelper {
 					{ share_link: shareLink },
 					tenantCode
 				)
+			}
+			try {
+				await cacheHelper.sessions.delete(tenantCode, sessionId)
+			} catch (cacheError) {
+				// Cache invalidation failure - continue operation
 			}
 			return responses.successResponse({
 				message: 'SESSION_LINK_GENERATED_SUCCESSFULLY',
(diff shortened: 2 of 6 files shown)


rakeshSgr:
@coderabbitai full review

src/services/sessions.js
```diff
@@ -1428,8 +1424,10 @@ module.exports = class SessionsHelper {
 				}
 			}
 
-			// Invalidate session cache after successful update
-			await cacheHelper.sessions.delete(tenantCode, sessionId)
+			try {
+				// Invalidate session cache after successful update
+				await cacheHelper.sessions.delete(tenantCode, sessionId)
+			} catch (error) {}
```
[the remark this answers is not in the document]
rakeshSgr: this is for update, not for delete

src/requests/user.js
```diff
@@ -950,7 +950,7 @@ const getUserDetailedListUsingCache = async function (usersMap, tenantCode, dele
 				usersInfo.missingUserIds,
 				options,
 				tenantCode,
-				unscopped
+				true
```
nevil-mathew: If we're hardcoding true, the function parameter doesn’t serve any purpose. We should either consistently use the value passed through the parameter or remove the parameter altogether to avoid confusion and potential issues.

src/services/sessions.js
```diff
@@ -690,7 +689,9 @@ module.exports = class SessionsHelper {
 			}
 
 			// Use direct database query instead of cache
```
nevil-mathew: Can we update or remove this invalid comment?

src/services/sessions.js
```diff
@@ -2278,27 +2281,26 @@ module.exports = class SessionsHelper {
 		try {
 			let email
 			let name
-			let userId
+			let userId = userTokenData?.user_id || userTokenData?.id
 			let emailTemplateCode = process.env.MENTEE_SESSION_CANCELLATION_EMAIL_TEMPLATE
 			// If mentee request unenroll get email and name from user service via api call.
 			// Else it will be available in userTokenData
 			if (isSelfUnenrollment) {
 				const userDetails = await mentorExtensionQueries.getMentorExtension(
```
nevil-mathew: Is it possible to get the mentor details from cache?

src/services/sessions.js
```diff
@@ -2476,6 +2475,11 @@ module.exports = class SessionsHelper {
 					tenantCode
 				)
 			}
+			try {
```
nevil-mathew: Why do we need to delete the cache if db is not updated? delete should be inside `if (!shareLink) {` block right?

src/services/sessions.js
```diff
@@ -2740,12 +2752,11 @@ module.exports = class SessionsHelper {
 
 	static async completed(sessionId, isBBB, tenantCode, orgCode) {
```
nevil-mathew, line 2758: We should invalidate the session cache after this DB update. If the cache TTL is longer than the session end time, the cached status can become stale.

src/services/sessions.js
nevil-mathew: `sessionDetail` variable in invalid, please update it to the correct variable.

src/services/sessions.js
```diff
@@ -4141,13 +4181,25 @@ module.exports = class SessionsHelper {
...
+					for (const userSession of removedSessionsDetail) {
+						try {
+							await cacheHelper.sessions.delete(tenantCode, userSession.id)
+						} catch (cacheError) {
+							console.error(`Cache deletion failed for session ${userSession.id}:`, cacheError)
+						}
+					}
+				}
+
 				await adminService.unenrollAndNotifySessionAttendees(
```
nevil-mathew, line 4211: We’re deleting the session cache and the mentor’s user cache, but not clearing attendee-side caches. If enrolled sessions count etc are cached in user details, this can lead to stale data.

nevil-mathew (changes requested):

src/requests/user.js
rakeshSgr: its fixed in other pr.

src/services/sessions.js
rakeshSgr: fixed3

src/services/sessions.js
rakeshSgr: done

src/services/sessions.js
rakeshSgr: fixed

PR #1529, opened 2026-01-12
Delete User Fix

sumanvpacewisdom:
## Release Notes

- **Attendee DTO Module**: Introduced new `src/dtos/attendee.js` module that exports an `attendeeData()` function for mapping attendee information to a standardized data transfer object with fields: attendee_id, enrolled_type, attendee_meeting_info, joined_at, and mentee_id.

- **Multi-tenant/Organization Notification Handling**: Enhanced `src/services/admin.js` to support array-based organization codes and tenant codes for notifications, enabling better handling of multi-tenant contexts. Updated `notifyAttendeesAboutSessionDeletion()` and `NotifySessionRequestedUsers()` method signatures to accept code arrays instead of single values, with internal logic normalizing to scalar values for database queries.

- **Notification Template Caching**: Improved notification template retrieval logic to normalize tenant and organization codes when they arrive as arrays, extracting the first element for consistent template lookups and cache operations.

- **Code Formatting**: Minor formatting improvements in `src/services/connections.js` with no functional changes.

| sumanvpacewisdom | 2757 | 0 |

### DIFF
--- a/src/services/admin.js
+++ b/src/services/admin.js
@@ -43,7 +43,12 @@ class NotificationHelper {
 				return true
 			}
 
-			const template = await cacheHelper.notificationTemplates.get(tenantCode, orgCode, templateCode)
+			// Handle arrays: extract first value (user codes), cacheHelper will fallback to defaults if not found
+			const userTenantCode = Array.isArray(tenantCode) ? tenantCode[0] : tenantCode
+			const userOrgCode = Array.isArray(orgCode) ? orgCode[0] : orgCode
+
+			// cacheHelper.get will automatically search user codes first, then defaults when cache is disabled
+			const template = await cacheHelper.notificationTemplates.get(userTenantCode, userOrgCode, templateCode)
 			if (!template) {
 				console.log(`Template ${templateCode} not found`)
 				return true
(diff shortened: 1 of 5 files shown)


nevil-mathew:
@coderabbitai full review

src/services/admin.js
```diff
@@ -222,10 +235,14 @@ module.exports = class AdminService {
...
-					attributes: ['user_id', 'name', 'email'],
+				const connectedMentors = await userExtensionQueries.getUsersByUserIds(
+					mentorIds,
+					{
+						attributes: ['user_id', 'name', 'email'],
+					},
 					tenantCode,
-				})
+					false
+				)
```
nevil-mathew: Why do we need to pass a new fn param here? Doesn't userExtensionQueries.getUsersByUserIds already have a default for this? Same for all other occurrences.

nevil-mathew (changes requested):

src/services/admin.js
sumanvpacewisdom: yes correct , I have changed in all the places , please check again now

PR #1550, opened 2026-01-22
Materialized views fixes

sumanvpacewisdom:
## Release Notes

- Multi-tenant support for materialized view refresh: admin endpoint accepts path parameter format `{tenantCode|modelName}` to refresh a single tenant or, when omitted, discovers and refreshes all tenants.
- Per-tenant result aggregation: consolidated response includes tenantsProcessed and per-tenant success/error details.
- Improved validation and error handling: parameter checks, model-existence checks, richer success/failure result objects, and clearer logging across refresh flow.
- Skip active refreshes: if an active REFRESH MATERIALIZED VIEW query exists, the refresh is skipped and logged with a skipped:true response.
- Scheduler robustness: startup validation for critical environment variables, encoded URL path usage for tenant and model parameters, stronger input/response validation, and more descriptive error messages.
- Process and job tracking: clearer logging and summaries for created scheduler jobs and overall operation outcomes.

### Contributor Statistics

| Author | Files Changed | Additions | Deletions |
|--------|---------------|-----------|-----------|
| sumanvpacewisdom | 3 | 166 | 41 |

### DIFF
--- a/src/controllers/v1/admin.js
+++ b/src/controllers/v1/admin.js
@@ -96,12 +96,12 @@ module.exports = class admin {
 			let modelName = null
 
 			// Check if tenant_code and model_name are encoded in path parameter (id)
-			// Format: {tenantCode|modelName}
+			// Format: {tenantCode|modelName} - Express auto-decodes URL params
 			if (req.params.id) {
 				const parts = req.params.id.split('|')
 				if (parts.length === 2) {
-					tenantCode = decodeURIComponent(parts[0])
-					modelName = decodeURIComponent(parts[1])
+					tenantCode = parts[0]
+					modelName = parts[1]
 				}
 			}
 
--- a/src/controllers/v1/admin.js
+++ b/src/controllers/v1/admin.js
@@ -104,60 +104,58 @@ module.exports = class admin {
 					modelName = decodeURIComponent(parts[1])
 				}
 			}
-			// Fallback to query params (for manual GET requests) or body (for POST requests)
-			if (!tenantCode) {
-				tenantCode = req.query.tenant_code || req.body?.tenant_code
-				modelName = req.query.model_name || req.body?.model_name
-			}
 
 			// If tenantCode is provided, refresh for that specific tenant
 			if (tenantCode) {
 				return await adminService.triggerPeriodicViewRefreshInternal(modelName, tenantCode)
-			}
+			} else {
+				// If no tenantCode provided, fetch all tenants dynamically and refresh for each
+				const tenants = await userExtensionQueries.getDistinctTenantCodes()
+				if (!tenants || tenants.length === 0) {
+					return responses.successResponse({
+						statusCode: httpStatusCode.ok,
+						message: 'NO_TENANTS_FOUND',
+						result: { tenantsProcessed: 0 },
+					})
+				}
 
-			// If no tenantCode provided, fetch all tenants dynamically and refresh for each
-			const tenants = await userExtensionQueries.getDistinctTenantCodes()
-			if (!tenants || tenants.length === 0) {
+				// Process each tenant
+				const results = []
+				for (const tenant of tenants) {
+					const tenantCodeToProcess = tenant.code
+					// Skip invalid tenant codes
+					if (!tenantCodeToProcess || tenantCodeToProcess === 'undefined') {
+						continue
+					}
+					try {
+						const result = await adminService.triggerPeriodicViewRefreshInternal(
+							modelName,
+							tenantCodeToProcess
+						)
+						results.push({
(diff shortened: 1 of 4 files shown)


src/controllers/v1/admin.js
```diff
@@ -92,23 +92,79 @@ module.exports = class admin {
...
+				if (parts.length === 2) {
+					tenantCode = decodeURIComponent(parts[0])
+					modelName = decodeURIComponent(parts[1])
 				}
+			}
+			// Fallback to query params (for manual GET requests) or body (for POST requests)
+			if (!tenantCode) {
+				tenantCode = req.query.tenant_code || req.body?.tenant_code
+				modelName = req.query.model_name || req.body?.model_name
+			}
```
nevil-mathew: Remove this

src/controllers/v1/admin.js
```diff
@@ -92,23 +92,79 @@ module.exports = class admin {
...
 
+			// If tenantCode is provided, refresh for that specific tenant
+			if (tenantCode) {
+				return await adminService.triggerPeriodicViewRefreshInternal(modelName, tenantCode)
+			}
+
+			// If no tenantCode provided, fetch all tenants dynamically and refresh for each
+			const tenants = await userExtensionQueries.getDistinctTenantCodes()
+			if (!tenants || tenants.length === 0) {
```
nevil-mathew: Why is this not inside else case? Or we can remove tthis, since we should/will always have the tenant code.

src/controllers/v1/admin.js
sumanvpacewisdom: Done

src/controllers/v1/admin.js
sumanvpacewisdom: Added else part

src/controllers/v1/admin.js
```diff
@@ -92,23 +92,77 @@ module.exports = class admin {
...
-				if (tenants.length > 0) {
-					return await adminService.triggerPeriodicViewRefreshInternal(req.query.model_name, tenants[0].code)
+			// Check if tenant_code and model_name are encoded in path parameter (id)
+			// Format: {tenantCode|modelName}
+			if (req.params.id) {
+				const parts = req.params.id.split('|')
+				if (parts.length === 2) {
+					tenantCode = decodeURIComponent(parts[0])
+					modelName = decodeURIComponent(parts[1])
+				}
```
nevil-mathew, line 105: Is this correct? We need to encode the URI while creating the job, right? For example, `/default|sessions` becomes `%2Fdefault%7Csessions`, so the `params.id` will not contain `|`, which would otherwise cause it to break, right?

sumanvpacewisdom: @nevil-mathew so we will need to change it to query params

src/scripts/viewsScript.js
```diff
@@ -131,34 +174,33 @@ const triggerPeriodicViewRefresh = async () => {
...
-					true,
-					mentoringBaseurl +
-						`/mentoring/v1/admin/triggerPeriodicViewRefreshInternal?model_name=${model}&tenant_code=${tenantCode}`,
-					globalOffset
-				)
+				// Encode tenant_code and model_name in the URL path using :id parameter
+				// Format: /mentoring/v1/admin/triggerPeriodicViewRefreshInternal/{tenantCode|modelName}
+				// The route pattern :version/:controller/:method/:id supports this
+				const encodedParams = `${encodeURIComponent(tenantCode)}|${encodeURIComponent(model)}`
+				const url = `${mentoringBaseurl}/mentoring/v1/admin/triggerPeriodicViewRefreshInternal/${encodedParams}`
```
nevil-mathew: `|` is not URL-safe in path segments. `encodeURIComponent()` doesn’t encode it when used separately.
Please encode the **entire segment** instead:

```js
const encodedParams = encodeURIComponent(`${tenantCode}|${model}`)
```

This avoids routing/proxy issues and makes the URL safe.
Make changes if used elsewhere.

nevil-mathew: @coderabbitai check if we have unsafe URL elsewhere in this script.

src/controllers/v1/admin.js
sumanvpacewisdom: Done

src/scripts/viewsScript.js
sumanvpacewisdom: Done

src/controllers/v1/admin.js
nevil-mathew: @coderabbitai check if the changes are applied.

PR #1559, opened 2026-02-10
undid location delete code

borkarsaish65:


src/services/sessions.js
```diff
@@ -1016,11 +1016,14 @@ module.exports = class SessionsHelper {
...
 				{ organization_id: mentorExtension.organization_id },
-				{ attributes: ['name'] }
+				{ attributes: ['name', 'organization_id'] }
 			)
 
 			if (orgDetails && orgDetails.name) {
-				sessionDetails.organization = orgDetails.name
+				sessionDetails['organization'] = {
+					id: mentorExtension.organization_id,
+					name: orgDetails.name,
```
rakeshSgr, line 1025: @borkarsaish65 do we not require code here ?

src/helpers/getEnrolledMentees.js
```diff
@@ -90,7 +90,6 @@ exports.getEnrolledMentees = async (sessionId, queryParams, userID) => {
 			'meta',
 			'email_verified',
 			'gender',
-			'location',
```
rakeshSgr, line 93: @borkarsaish65 why this is removed ?

src/services/sessions.js
borkarsaish65: other api response does not contain code anymore, these 2 fields are provided

src/helpers/getEnrolledMentees.js
borkarsaish65: location which is newly added is getting deleted due to this code, vishwa wanted location in the response so 
made the code change.

PR #1577, opened 2026-02-17
User mentor extension update

sumanvpacewisdom:
### Key Changes

- **Partition Key Protection**: Updated `updateMentorExtension` and `updateRoleExtension` functions to remove `tenant_code` from update payloads, preventing accidental updates to Citus partition keys
- **Data Integrity**: Modified `removeMentorDetails` to exclude `organization_code` and `tenant_code` from nullification to prevent NOT NULL constraint violations
- **Session Query Enhancement**: Updated `getSessionsAssignedToMentor` to accept `tenantCode` parameter and properly filter results by tenant
- **Circular Dependency Resolution**: Introduced new `organizationCache` wrapper module to break circular dependency between cacheHelper and user request handlers
- **Admin Service Corrections**: Fixed multiple issues in admin service including:
  - Corrected `getSessionsMapping` calls to pass three arguments instead of two
  - Removed invalid join with non-existent `request_session_mapping` table
  - Updated notification parameter naming to use singular `tenantCode` instead of `tenantCodes`
  - Improved data flow for user detail retrieval and session mapping aggregation
- **User Extension Protection**: Updated `updateMenteeExtension` to remove partition key columns (`user_id`, `tenant_code`) before update operations

### Lines Changed by Author

| sumanvpacewisdom | 116 | 31 |

### DIFF
--- a/src/database/queries/roleExtentions.js
+++ b/src/database/queries/roleExtentions.js
@@ -76,6 +76,10 @@ module.exports = class RoleExtensionService {
 
 	static async updateRoleExtension(title, updateData, tenantCode) {
 		try {
+			// Remove partition key columns - Citus doesn't allow updating these
+			if (updateData.tenant_code) {
+				delete updateData['tenant_code']
+			}
 			const filter = { title: title, tenant_code: tenantCode }
 
 			const [rowsUpdated, updatedExtension] = await RoleExtension.update(updateData, {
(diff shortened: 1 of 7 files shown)


src/database/queries/roleExtentions.js
```diff
@@ -76,6 +76,10 @@ module.exports = class RoleExtensionService {
 
 	static async updateRoleExtension(title, updateData, tenantCode) {
 		try {
+			// Remove partition key columns - Citus doesn't allow updating these
+			if (updateData.tenant_code) {
+				delete updateData['tenant_code']
+			}
```
nevil-mathew, line 82: This will be removed in the validator right if added in the req body?

PR #1607, opened 2026-03-23
Cache issues related to entity_type , entity and notifications 

sumanvpacewisdom:
## Release Notes

- **Cache alignment in cacheHelper**: Updated session and mentee caching methods to use correct parameter signatures (`tenantCode`, `id`, `data` instead of including `organizationCode` as the second argument)
- **Entity type cache invalidation**: Refined cache deletion logic to target specific model + value entries instead of broad namespace keys
- **Simplified entity caching**: Removed read-through caching logic and associated cache invalidation during entity CRUD operations
- **Notification template cache optimization**: Refactored cache invalidation to capture pre-update template code values, eliminating redundant post-update queries and improving cache invalidation accuracy

## Lines Changed by Author

| Author | Added | Removed | Net |
|--------|-------|---------|-----|
| sumanvpacewisdom | 29 | 93 | -64 |

### DIFF
--- a/src/generics/cacheHelper.js
+++ b/src/generics/cacheHelper.js
@@ -484,7 +484,7 @@ const sessions = {
 
 					if (sessionDetails) {
 						// Cache the session data directly
-						await this.set(tenantCode, organizationCode, sessionId, sessionDetails)
+						await this.set(tenantCode, sessionId, sessionDetails)
 						dbFetchedSessions.push(sessionDetails)
 						console.log(`✅ [getSessionKafka] Session ${sessionId} fetched and cached`)
 					}
@@ -1322,7 +1322,7 @@ const mentee = {
 					if (usersFromDb && usersFromDb.length > 0) {
 						// Cache each fetched user individually
 						for (const user of usersFromDb) {
-							await this.set(tenantCode, organizationCode, user.user_id, user)
+							await this.set(tenantCode, user.user_id, user)
 						}
 
 						// Add fetched users to result
--- a/src/services/entity-type.js
+++ b/src/services/entity-type.js
@@ -289,11 +289,12 @@ module.exports = class EntityHelper {
 			}
 
 			// Clear cache for affected models before deletion
-			await this._clearUserCachesForEntityTypeChange(organizationCode, tenantCode, {
-				id: entityToDelete.id,
-				value: entityToDelete.value,
-				modelNames: entityToDelete.model_names,
-			})
+			await this._clearUserCachesForEntityTypeChange(
+				organizationCode,
+				tenantCode,
+				entityToDelete.model_names ? entityToDelete.model_names[0] : null,
+				entityToDelete.value
+			)
 
 			// SECOND: Delete from database
 			const deleteCount = await entityTypeQueries.deleteOneEntityType(id, organizationCode, tenantCode)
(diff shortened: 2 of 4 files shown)


sumanvpacewisdom:
@coderabbitai full review

src/services/entity-type.js
```diff
@@ -289,11 +289,12 @@ module.exports = class EntityHelper {
...
 			// Clear cache for affected models before deletion
-			await this._clearUserCachesForEntityTypeChange(organizationCode, tenantCode, {
-				id: entityToDelete.id,
-				value: entityToDelete.value,
-				modelNames: entityToDelete.model_names,
-			})
+			await this._clearUserCachesForEntityTypeChange(
+				organizationCode,
+				tenantCode,
+				entityToDelete.model_names ? entityToDelete.model_names[0] : null,
```
nevil-mathew, line 295: Why are we using just the first element from the array? What happens if there are two models for the same entity type?

src/services/entity-type.js
```diff
@@ -544,13 +545,13 @@ module.exports = class EntityHelper {
 				})
 			)
 
-			// 2. Clear entity type caches for unified model strategy
+			// 2. Clear entity type cache for the specific model + value
 			if (modelName) {
 				clearPromises.push(
 					cacheHelper.entityTypes
-						.delete(tenantCode, organizationCode, `model:${modelName}:__ALL__`)
+						.delete(tenantCode, organizationCode, modelName, entityValue)
```
nevil-mathew, line 552: I’m assuming `__ALL__` is no longer being used in the cache keys?

nevil-mathew (changes requested):

nevil-mathew (commented):
While you’ve fixed the issue in cacheHelper.js line 487, it still persists in 442 and 511. Shouldn’t we fix this as well? Are there any other occurrences that haven’t been updated?

nevil-mathew:
Similar issue in line 1283, 1352

src/services/entity-type.js
```diff
@@ -289,11 +289,12 @@ module.exports = class EntityHelper {
 			}
 
 			// Clear cache for affected models before deletion
-			await this._clearUserCachesForEntityTypeChange(organizationCode, tenantCode, {
-				id: entityToDelete.id,
-				value: entityToDelete.value,
-				modelNames: entityToDelete.model_names,
-			})
+			await this._clearUserCachesForEntityTypeChange(
```
nevil-mathew, line 292: Why are we calling _clearUserCachesForEntityTypeChange twice(line 343)?

src/generics/cacheHelper.js
nevil-mathew, line 1333: Is `userIsMentorMap` declared in this fn?

nevil-mathew, line 1: `getMentorKafka` (src/generics/cacheHelper.js:1036) still has the same `.get()` parameter mismatch.

It calls `this.get(tenantCode, organizationCode, userIds)`, but `mentor.get` (line 884) only accepts `(tenantCode, mentorId)`. The fallback at line 1105 has the same issue. The PR fixed this for session and mentee, but not here. No impact for now since it’s not used.

Also, in `getMenteeKafka` (src/generics/cacheHelper.js:1332–1333), `userIsMentorMap` is referenced but never defined. This throws a `ReferenceError` (caught at line 1337), causing errors to be logged every time DB users are found, skipping the success log at 1335, and also shadowing `userMentorEntries` from line 1311.

nevil-mathew (changes requested):

src/services/entity-type.js
sumanvpacewisdom: We are not using all anywhere correct

PR #1610, opened 2026-03-25
Feature/cach fix tenant

borkarsaish65:
## Release Notes

- Removed cache invalidation for entity list caching after entity create, update, and delete operations, keeping only entityType cache invalidation intact
- Removed read-path caching logic in entity list method; pagination cache lookup/set for `entity_list` was deleted and the method now always fetches results directly
- Removed cache-first logic in modules list method that conditionally retrieved cached `modules_list` results; pagination/search behavior is now independent of cached data
- Updated session details caching approach to cache raw session details instead of processed response; `processDbResponse` computation moved to occur after the caching block

## Author Contributions

| borkarsaish65 | 21 | 123 |

### DIFF
--- a/src/services/sessions.js
+++ b/src/services/sessions.js
@@ -1757,11 +1757,9 @@ module.exports = class SessionsHelper {
 
 			const validationData = removeDefaultOrgEntityTypes(sessionEntityTypes, defaults.orgCode)
 
-			const processDbResponse = utils.processDbResponse(sessionDetails, validationData)
-
 			if (utils.isNumeric(id)) {
 				try {
-					const cacheCopy = { ...processDbResponse }
+					const cacheCopy = { ...sessionDetails }
 					delete cacheCopy.is_enrolled
 					delete cacheCopy.enrolment_type
 					await cacheHelper.sessions.set(tenantCode, cacheCopy.id, cacheCopy)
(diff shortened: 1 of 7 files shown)


nevil-mathew:
@coderabbitai full review

src/services/sessions.js
```diff
@@ -1757,11 +1757,9 @@ module.exports = class SessionsHelper {
 
 			const validationData = removeDefaultOrgEntityTypes(sessionEntityTypes, defaults.orgCode)
 
-			const processDbResponse = utils.processDbResponse(sessionDetails, validationData)
-
 			if (utils.isNumeric(id)) {
 				try {
-					const cacheCopy = { ...processDbResponse }
+					const cacheCopy = { ...sessionDetails }
```
nevil-mathew, line 1762: Why are we storing the raw response in the cache? Has the cache hit flow been updated to handle this new logic? What’s the reason behind this change?

borkarsaish65: The change was needed because processDbResponse() transforms type into an object { label: "PUBLIC", value: "PUBLIC" }. When the processed result was cached, it was breaking the session update and delete API flows — the comparison sessionDetail.type != bodyData.type was always true (object vs string), triggering CANNOT_EDIT_MENTOR_AND_TYPE incorrectly.

Storing raw data fixes that comparison. The cache-hit path in the details endpoint has been verified and works correctly with the raw data.

PR #1611, opened 2026-03-26
Add migration to create indexes for tenant-specific materialized view…

sumanvpacewisdom:
…s and update index query handling

- Introduced a new migration script to create indexes on tenant-specific materialized views for sessions and user extensions.
- Updated index query handling in materializedViews.js to accept tenantCode as a parameter.
- Refactored mViewsIndexQueries.js to generate tenant-specific index queries dynamically.

## Release Notes

- Add Sequelize migration to create tenant-specific indexes on materialized views for sessions (`${tenantCode}_m_sessions`) and user extensions (`${tenantCode}_m_user_extensions`)
- Indexes for sessions include:
  - composite index on (mentor_organization_id, status, type, mentor_id)
  - composite index on (mentor_id, status, start_date)
  - composite index on (status, type, start_date)
- Indexes for user extensions include:
  - conditional index on lower(name) WHERE is_mentor = true (scoped to organization_id)
  - index on email
  - index on organization_code
- Migration `up` creates indexes using `CREATE INDEX IF NOT EXISTS` for each active tenant; `down` drops the tenant-specific indexes using `DROP INDEX IF EXISTS`
- Refactor to generate tenant-scoped index SQL:
  - replace static index list with `getIndexQueries(tenantCode)` in `mViewsIndexQueries.js`
  - update `materializedViews.js` to pass `tenantCode` into index execution (`executeIndexQueries(modelName, tenantCode)`)
- Error handling added around migration operations (logs and rethrows on failure)

## Contribution Summary

| sumanvpacewisdom | 94 | 10 |


src/generics/mViewsIndexQueries.js
```diff
@@ -1,14 +1,20 @@
-const indexQueries = [
+const getIndexQueries = (tenantCode) => [
 	{
 		modelName: 'Session',
 		queries: [
-			`CREATE INDEX IF NOT EXISTS idx_filtered_sessions ON m_sessions (mentor_organization_id, status, type, mentor_id);`,
+			`CREATE INDEX IF NOT EXISTS ${tenantCode}_idx_filtered_sessions ON ${tenantCode}_m_sessions (mentor_organization_id, status, type, mentor_id);`,
```
rakeshSgr, line 5: @sumanvpacewisdom is tenant_code is not require ?

src/generics/mViewsIndexQueries.js
```diff
@@ -1,14 +1,20 @@
...
+			`CREATE INDEX IF NOT EXISTS ${tenantCode}_idx_sessions_mentor_status_date ON ${tenantCode}_m_sessions (mentor_id, status, start_date);`,
+			`CREATE INDEX IF NOT EXISTS ${tenantCode}_idx_sessions_status_type_date ON ${tenantCode}_m_sessions (status, type, start_date);`,
 		],
 	},
 	{
-		modelName: 'MentorExtension',
-		queries: [`CREATE INDEX IF NOT EXISTS idx_m_mentor_extensions_email ON m_mentor_extensions (email);`],
+		modelName: 'UserExtension',
+		queries: [
+			`CREATE INDEX IF NOT EXISTS ${tenantCode}_idx_user_ext_org_name ON ${tenantCode}_m_user_extensions (organization_id, lower(name)) WHERE is_mentor = true;`,
```
rakeshSgr, line 13: @sumanvpacewisdom tenant_code is not required ?

src/generics/mViewsIndexQueries.js
sumanvpacewisdom: @rakeshSgr 
tenant_code is not required here because these are per-tenant materialized views ${tenantCode}_m_sessions already contains only that tenant's data. 
Adding tenant_code to the index would be redundant since it has the same value for every row in the table. tenant_code in indexes is only needed on shared base tables where all tenants' data coexists.

src/generics/mViewsIndexQueries.js
sumanvpacewisdom: @rakeshSgr tenant_code is not required here because these are per-tenant materialized views ${tenantCode}_m_sessions already contains only that tenant's data. 
Adding tenant_code to the index would be redundant since it has the same value for every row in the table. tenant_code in indexes is only needed on shared base tables where all tenants' data coexists.

PR #1612, opened 2026-03-26
Refactor cache handling in session and user retrieval methods across …

sumanvpacewisdom:
…services to improve efficiency. Updated cache deletion logic in AdminService and streamlined user cache clearing in EntityHelper. Enhanced organization extension updates in OrganizationService by including tenantCode in the upsert method.

## Release Notes

- Cache key optimization: removed redundant organizationCode from single-ID cache lookups in getSessionKafka, getMentorKafka, and getMenteeKafka to improve cache efficiency.
- Simplified cache invalidation: removed redundant post-mutation invalidations and entity-list cache operations; capture old template code once and only delete cache for the old code in NotificationTemplateHelper.
- Broader entity-type handling: entity-type updates and deletes now iterate all model names and clear caches per model value after successful DB deletion.
- Tenancy support: pass tenantCode into organisationExtensionQueries.upsert to ensure tenant-scoped organization extension updates.
- AdminService cache deletion fix: delete mentee cache using tenantCode + menteeId (removed organization_code) and corrected error logging to reference mentee_id.
- Local refactors: adjusted variable mutability and simplified mentor/mentee mapping logic in cacheHelper methods.

### Lines Changed by Author

| Author | Added | Removed |
|--------|:-----:|:-------:|
| sumanvpacewisdom | 37 | 132 |

### DIFF
--- a/src/generics/cacheHelper.js
+++ b/src/generics/cacheHelper.js
@@ -439,7 +439,7 @@ const sessions = {
 		try {
 			// Handle single session ID case
 			if (typeof sessionIds === 'string') {
-				return await this.get(tenantCode, organizationCode, sessionIds)
+				return await this.get(tenantCode, sessionIds)
 			}
 
 			// Handle array case - smart caching with immediate DB fetch
@@ -508,7 +508,7 @@ const sessions = {
 			const fallbackResults = []
 			for (const sessionId of Array.isArray(sessionIds) ? sessionIds : [sessionIds]) {
 				try {
-					const session = await this.get(tenantCode, organizationCode, sessionId)
+					const session = await this.get(tenantCode, sessionId)
 					if (session) fallbackResults.push(session)
 				} catch (fallbackError) {
 					console.error(
@@ -1033,7 +1033,7 @@ const mentor = {
 		try {
 			// Handle single user ID case
 			if (typeof userIds === 'string') {
-				return await this.get(tenantCode, organizationCode, userIds)
+				return await this.get(tenantCode, userIds)
 			}
 
 			// Handle array case - smart caching with immediate DB fetch
@@ -1061,7 +1061,7 @@ const mentor = {
 
 			// Step 4: Fetch missing users from DB immediately using getUsersByUserIds
 			const dbFetchedUsers = []
-			const userMentorEntries = []
+			let userMentorEntries = []
 			if (missingUserIds.length > 0) {
 				try {
 					const usersFromDb = await userQueries.getUsersByUserIds(missingUserIds, {}, tenantCode, false)
@@ -1102,7 +1102,7 @@ const mentor = {
 			const fallbackResults = []
 			for (const userId of Array.isArray(userIds) ? userIds : [userIds]) {
 				try {
-					const user = await this.get(tenantCode, organizationCode, userId)
+					const user = await this.get(tenantCode, userId)
 					if (user) fallbackResults.push(user)
 				} catch (fallbackError) {
 					console.error(`❌ [getMenteeKafka] Fallback failed for user ${userId}:`, fallbackError.message)
@@ -1280,7 +1280,7 @@ const mentee = {
 		try {
 			// Handle single user ID case
 			if (typeof userIds === 'string') {
-				return await this.get(tenantCode, organizationCode, userIds)
+				return await this.get(tenantCode, userIds)
 			}
 
 			// Handle array case - smart caching with immediate DB fetch
@@ -1308,7 +1308,6 @@ const mentee = {
 
 			// Step 3: Fetch missing users from DB immediately using getUsersByUserIds
 			const dbFetchedUsers = []
-			let userMentorEntries = []
 			if (missingUserIds.length > 0) {
 				try {
 					console.log(`🔄 [getMenteeKafka] Fetching ${missingUserIds.length} users from database`)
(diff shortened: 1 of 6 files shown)


src/services/entity-type.js
```diff
@@ -289,12 +290,18 @@ module.exports = class EntityHelper {
...
+				await this._clearUserCachesForEntityTypeChange(
+					organizationCode,
+					tenantCode,
+					modelName,
+					entityToDelete.value
+				)
+			}
+			if (deleteModelNames.length === 0) {
+				await this._clearUserCachesForEntityTypeChange(organizationCode, tenantCode, null, entityToDelete.value)
+			}
```
nevil-mathew: Why are we deleting cache if model names are empty? model_names can never be empty, right? And cache clearing should be after db op.

src/services/entity-type.js
```diff
@@ -118,12 +118,13 @@ module.exports = class EntityHelper {
...
-				tenantCode,
-				updatedEntity.model_names ? updatedEntity.model_names[0] : null,
-				updatedEntity.value
-			)
+			const updatedModelNames = updatedEntity.model_names || []
+			for (const modelName of updatedModelNames) {
+				await this._clearUserCachesForEntityTypeChange(orgCode, tenantCode, modelName, updatedEntity.value)
+			}
+			if (updatedModelNames.length === 0) {
+				await this._clearUserCachesForEntityTypeChange(orgCode, tenantCode, null, updatedEntity.value)
```
nevil-mathew: Why are we deleting cache if model names are empty? model_names can never be empty, right?

src/services/admin.js
```diff
@@ -694,7 +694,7 @@ module.exports = class AdminService {
 			if (unenrollDetails && unenrollDetails.deletedCount > 0 && Array.isArray(unenrollDetails.deletedRecords)) {
 				for (const menteeData of unenrollDetails.deletedRecords) {
 					try {
-						await cacheHelper.mentee.delete(tenantCode, menteeData.organization_code, menteeData.mentee_id)
+						await cacheHelper.mentee.delete(tenantCode, menteeData.mentee_id)
 					} catch (cacheError) {
 						console.error(`Cache deletion failed for mentee ${menteeId}:`, cacheError)
```
nevil-mathew: Where is mentee_id declared? Shouldn't it be menteeData.mentee_id?

src/generics/cacheHelper.js
nevil-mathew, line 1: Why do we have two `_sendToKafkaBackground` functions? Which one is actually correct? This creates unnecessary ambiguity. They take different parameters but essentially do the same thing—what’s the reason for that? Were these changes properly tested and verified?

Also, there’s a major issue: the entire Kafka background cache warming path doesn’t seem to work. In `kafka-communication.js`, only two functions are exported—`pushEmailToKafka` and `clearInternalCache`. There’s no `pushToKafka` export at all.

So how was this tested?

src/generics/cacheHelper.js
```diff
@@ -1308,7 +1308,6 @@ const mentee = {
 
 			// Step 3: Fetch missing users from DB immediately using getUsersByUserIds
 			const dbFetchedUsers = []
-			let userMentorEntries = []
 			if (missingUserIds.length > 0) {
```
nevil-mathew, line 1323: The getMenteeKafka method has no equivalent call to this._sendToKafkaBackground() to getMentorKafka. What is the use of getMenteeKafka?

src/services/admin.js
sumanvpacewisdom: Correct , menteeId was not declared in that scope , the loop variable is menteeData. Fixed the error log to use menteeData.mentee_id instead.

src/services/entity-type.js
sumanvpacewisdom: Yes Correct, model_names can never be empty since it's required when creating an entity type. That fallback was unnecessary and I have removed it. 
Also moved the cache clearing to happen after the DB delete, not before.

src/services/entity-type.js
sumanvpacewisdom: Same answer as above comment .

src/generics/cacheHelper.js
sumanvpacewisdom: This needs to be done later as discussion

src/generics/cacheHelper.js
sumanvpacewisdom: This needs to be done later as discussion

src/services/notification.js
```diff
@@ -84,23 +89,21 @@ module.exports = class NotificationTemplateHelper {
...
+						oldCode
 					)
 				}
-				if (existingTemplate?.code && existingTemplate.code !== templateCode) {
+				if (newCode && newCode !== oldCode) {
 					await cacheHelper.notificationTemplates.delete(
 						tenantCode,
 						tokenInformation.organization_code,
-						existingTemplate.code
+						newCode
```
nevil-mathew: Why are we deleting it with newCode? Is there any possibility that cache hit with just updated code?

src/services/entity.js
```diff
@@ -441,35 +395,13 @@ module.exports = class EntityHelper {
...
-					}
-				}
-			}
+			const entities = await entityQueries.getAllEntitiesWithEntityTypeDetails(
+				filter,
+				[tenantCode],
+				pageNo,
+				pageSize,
+				searchText
+			)
```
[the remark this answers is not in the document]
sumanvpacewisdom: It was not breaking anything , but it was expecting string , so i have reverted the changes to string

src/services/notification.js
sumanvpacewisdom: No we have no posibility of cache hit with just updated code , check and verified , hence removing the code .

Reason not caught in testing : it was added because it existed in the original code, tests passed because deleting a non-existent cache key silently does nothing.

PR #1613, opened 2026-03-27
fix: cache invalidation and org priority bugs in forms

borkarsaish65:
- Add cache invalidation on create to prevent stale fallback (default org) data
- Fix read() priority to prefer user's org over default org (was using tenant_code, should be organization_code)
- Fix readAllFormsVersion() to cache user org form over default org form when both exist

## Release Notes

- **Cache invalidation on form creation:** Added cache invalidation after successful form creation to prevent stale default organization data from being served
- **Fixed org priority in form read:** Updated `read()` method to prioritize user's organization form over default organization form by using `organization_code` instead of `tenant_code` for comparison
- **Fixed org priority in form version caching:** Updated `readAllFormsVersion()` to cache user organization forms over default organization forms when both exist, preferring custom organization entries

## Lines Changed by Author

| borkarsaish65 | 15 | 3 |

### DIFF
--- a/src/services/form.js
+++ b/src/services/form.js
@@ -29,7 +29,16 @@ module.exports = class FormsHelper {
 			bodyData['organization_code'] = orgCode
 			const form = await formQueries.createForm(bodyData, tenantCode, orgCode)
 
-			await KafkaProducer.clearInternalCache('formVersion')
+			//			await KafkaProducer.clearInternalCache('formVersion')
+
+			// Invalidate cache so stale fallback (default org) is not served
+			try {
+				if (bodyData.type && bodyData.sub_type) {
+					await cacheHelper.forms.delete(tenantCode, orgCode, bodyData.type, bodyData.sub_type)
+				}
+			} catch (cacheError) {
+				console.error('Failed to invalidate form cache:', cacheError)
+			}
 
 			return responses.successResponse({
 				statusCode: httpStatusCode.created,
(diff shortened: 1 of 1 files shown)


borkarsaish65:
@coderabbitai full review

src/services/form.js
```diff
@@ -31,6 +31,15 @@ module.exports = class FormsHelper {
...
 			//			await KafkaProducer.clearInternalCache('formVersion')
 
+			// Invalidate cache so stale fallback (default org) is not served
+			try {
+				if (bodyData.type && bodyData.sub_type) {
+					await cacheHelper.forms.delete(tenantCode, orgCode, bodyData.type, bodyData.sub_type)
+				}
+			} catch (cacheError) {
+				console.error('Failed to invalidate form cache:', cacheError)
+			}
```
nevil-mathew, line 41: Why do we need to delete the cache when a new form is created? It’s new, so it wouldn’t have been cached anyway. Are we caching a list somewhere? It doesn’t look like we’re clearing any list cache either.

src/services/form.js
```diff
@@ -240,7 +249,10 @@ module.exports = class FormsHelper {
 									)
```
nevil-mathew, line 249: Why are we caching this? We’re not even using cache for the list.

Here we’re caching individual forms, but `await form.getAllFormsVersion` only returns the version without the rest of the data. That means reads will fail, because the cache ends up serving incomplete, unusable form data.

nevil-mathew (changes requested):

src/services/form.js
nevil-mathew, line 118: It looks like we’re only clearing the cache for the current org. What happens to other orgs when default org data is updated (same for delete, etc.)? Since default org data is shared across all orgs in the tenant, shouldn’t we be invalidating their caches as well?

nevil-mathew (changes requested):

nevil-mathew:
OT
In `cacheHelper.forms.get` (cacheHelper.js), `formFromDb` is initialized as `null`, and although `formQueries.findFormsByFilter(...)` is called at line 748, its result is never assigned back to `formFromDb`. So `if (formFromDb)` will always evaluate to false.

PR #1614, opened 2026-03-27
fix: cache invalidation and priority bugs in notification templates

borkarsaish65:
- Add cache invalidation on create to prevent stale fallback (default org) data
- Fix update() to fetch template BEFORE the update for correct old-code cache invalidation
- Fix readAllNotificationTemplates() race condition where default org template could overwrite custom org template in cache

## Release Notes

- Added cache invalidation on create to prevent stale fallback (default organization) template data from being served
- Fixed `update()` to fetch the existing template before performing database updates, ensuring correct old-code cache invalidation when template codes change
- Fixed `readAllNotificationTemplates()` race condition where default organization templates could overwrite custom organization templates in cache by implementing template deduplication that prioritizes organization-scoped templates

## Contributions by Author

| borkarsaish65 | 30 | 19 |

### DIFF
--- a/src/services/notification.js
+++ b/src/services/notification.js
@@ -37,6 +37,17 @@ module.exports = class NotificationTemplateHelper {
 
 			const createdNotification = await notificationTemplateQueries.create(bodyData, tenantCode)
 
+			// Invalidate cache so stale fallback (default org) is not served
+			try {
+				await cacheHelper.notificationTemplates.delete(
+					tenantCode,
+					tokenInformation.organization_code,
+					bodyData.code
+				)
+			} catch (cacheError) {
+				console.error('❌ Failed to invalidate notification template cache:', cacheError)
+			}
+
 			return responses.successResponse({
 				statusCode: httpStatusCode.created,
 				message: 'NOTIFICATION_TEMPLATE_CREATED_SUCCESSFULLY',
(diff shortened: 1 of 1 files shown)


borkarsaish65:
@coderabbitai full review

src/services/notification.js
```diff
@@ -37,6 +37,17 @@ module.exports = class NotificationTemplateHelper {
...
+			// Invalidate cache so stale fallback (default org) is not served
+			try {
+				await cacheHelper.notificationTemplates.delete(
+					tenantCode,
+					tokenInformation.organization_code,
+					bodyData.code
+				)
+			} catch (cacheError) {
+				console.error('❌ Failed to invalidate notification template cache:', cacheError)
+			}
```
nevil-mathew, line 51: Why are we invalidating cache at create?

src/services/notification.js
```diff
@@ -228,21 +242,18 @@ module.exports = class NotificationTemplateHelper {
...
+							templatesByCode[template.code] = template
 						}
 					}
 
+					const cachePromises = Object.entries(templatesByCode).map(([code, template]) =>
+						cacheHelper.notificationTemplates.set(tenantCode, organizationCode, code, template)
+					)
 					await Promise.all(cachePromises)
 				} catch (cacheError) {
 					console.warn('Failed to cache individual notification templates:', cacheError)
```
nevil-mathew: Remove this.

Why do we have to cache here?

nevil-mathew (changes requested):

src/services/notification.js
borkarsaish65: removed now.

src/services/notification.js
borkarsaish65: for the following scenario:

Scenario: org ABC requests welcome_email template → their org doesn't have it → default org's template gets cached under ABC key due to fallback logic.

Now org ABC creates their own welcome_email template.

Without invalidation → next read still hits cache → returns default org's old template, not the one they just created.

With invalidation → cache key cleared on create → next read fetches their own new template from DB.

PR #1618, opened 2026-03-30
entities and entity type fix for cache invalidation

borkarsaish65:
## Release Notes

- Add cacheHelper.sessions.deleteAll(tenantCode, orgCode = null) to bulk-invalidate session caches scoped to an org or tenant-wide.
- Add cacheHelper.entityTypes.deleteEntityTypesAcrossAllOrgs(tenantCode, modelName, entityValue) to evict entity-type keys across all orgs in a tenant via Redis SCAN + delete.
- Add cacheHelper.mentor.deleteAll(tenantCode) and cacheHelper.mentee.deleteAll(tenantCode) to clear tenant-wide mentor/mentee profile caches.
- EntityHelper.update, EntityHelper.delete, and deleteEntityTypesAndEntities now compute whether the change is for the default organisation and pass an allOrgs flag to _clearUserCachesForEntityTypeChange to choose between org-scoped vs tenant-wide invalidation.
- _clearUserCachesForEntityTypeChange gained an allOrgs parameter; when true it:
  - uses cacheHelper.entityTypes.deleteEntityTypesAcrossAllOrgs(...) for entity-type cache sweeping,
  - uses cacheHelper.mentee.deleteAll/cacheHelper.mentor.deleteAll for mentee/mentor caches,
  - uses cacheHelper.sessions.deleteAll(tenantCode) for session caches.
- Stopped relying on updatedEntityType[0] for clearing user caches; cleanup decisions are driven by the original entity data and the allOrgs flag.
- getAllUsersByOrgId now computes the tenant view name before validating orgCodes so tenant-scoped logic can access the view name even when orgCodes is null/empty.
- Files changed: src/generics/cacheHelper.js, src/services/entity-type.js, src/database/queries/userExtension.js

## Contributions by Author

| Author | Added | Removed |
|--------|------:|--------:|
| borkarsaish65 | 105 | 97 |

### DIFF
--- a/src/database/queries/userExtension.js
+++ b/src/database/queries/userExtension.js
@@ -712,11 +712,12 @@ module.exports = class MenteeExtensionQueries {
 
 	static async getAllUsersByOrgId(orgCodes, tenantCode) {
 		try {
-			const viewName = utils.getTenantViewName(tenantCode, MenteeExtension.tableName)
-
 			if (!Array.isArray(orgCodes) || orgCodes.length === 0) {
 				return []
 			}
+
+			const viewName = utils.getTenantViewName(tenantCode, MenteeExtension.tableName)
+
 			const query = `
 			SELECT user_id
 			FROM ${viewName}
--- a/src/database/queries/userExtension.js
+++ b/src/database/queries/userExtension.js
@@ -712,11 +712,17 @@ module.exports = class MenteeExtensionQueries {
 
 	static async getAllUsersByOrgId(orgCodes, tenantCode) {
 		try {
+			const viewName = utils.getTenantViewName(tenantCode, MenteeExtension.tableName)
+
+			// null orgCodes = all orgs (used for default-org cross-org cache invalidation)
+			if (orgCodes === null) {
+				const query = `SELECT user_id FROM ${viewName} WHERE tenant_code = :tenantCode`
+				return await Sequelize.query(query, { type: QueryTypes.SELECT, replacements: { tenantCode } })
+			}
+
 			if (!Array.isArray(orgCodes) || orgCodes.length === 0) {
 				return []
 			}
-			const viewName = utils.getTenantViewName(tenantCode, MenteeExtension.tableName)
-
 			const query = `
 			SELECT user_id
 			FROM ${viewName}
--- a/src/generics/cacheHelper.js
+++ b/src/generics/cacheHelper.js
@@ -373,9 +373,7 @@ const sessions = {
 		const pattern = orgCode
 			? `tenant:${tenantCode}:org:${orgCode}:sessions:*`
 			: `tenant:${tenantCode}:org:*:sessions:*`
-		console.log(`[SessionCache] deleteAll - scanning and deleting pattern: ${pattern}`)
 		const result = await scanAndDelete(pattern)
-		console.log(`[SessionCache] deleteAll - deleted ${result} key(s)`)
 		return result
 	},
 
--- a/src/generics/cacheHelper.js
+++ b/src/generics/cacheHelper.js
@@ -948,9 +948,7 @@ const mentor = {
 
 	async deleteAll(tenantCode) {
 		const pattern = `tenant:${tenantCode}:mentor:*`
-		console.log(`[MentorCache] deleteAll - scanning and deleting pattern: ${pattern}`)
 		const result = await scanAndDelete(pattern)
-		console.log(`[MentorCache] deleteAll - deleted ${result} key(s)`)
 		return result
 	},
 
@@ -1195,9 +1193,7 @@ const mentee = {
 
 	async deleteAll(tenantCode) {
 		const pattern = `tenant:${tenantCode}:mentee:*`
-		console.log(`[MenteeCache] deleteAll - scanning and deleting pattern: ${pattern}`)
 		const result = await scanAndDelete(pattern)
-		console.log(`[MenteeCache] deleteAll - deleted ${result} key(s)`)
 		return result
 	},
 
--- a/src/generics/cacheHelper.js
+++ b/src/generics/cacheHelper.js
@@ -369,10 +369,8 @@ const sessions = {
 		return del(cacheKey, { useInternal })
 	},
 
-	async deleteAll(tenantCode, orgCode = null) {
-		const pattern = orgCode
-			? `tenant:${tenantCode}:org:${orgCode}:sessions:*`
-			: `tenant:${tenantCode}:org:*:sessions:*`
+	async deleteAll(tenantCode) {
+		const pattern = `tenant:${tenantCode}:sessions:*`
 		const result = await scanAndDelete(pattern)
 		return result
 	},
--- a/src/generics/cacheHelper.js
+++ b/src/generics/cacheHelper.js
@@ -369,6 +369,16 @@ const sessions = {
 		return del(cacheKey, { useInternal })
 	},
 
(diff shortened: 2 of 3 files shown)


borkarsaish65:
@coderabbitai full review

src/services/entity-type.js
```diff
@@ -107,9 +107,19 @@ module.exports = class EntityHelper {
...
+							// Default org update: other orgs may have cached this entity type via fallback
+							// under their own org key — sweep all of them
+							await cacheHelper.entityTypes.deleteEntityTypesAcrossAllOrgs(
+								tenantCode,
+								modelName,
+								originalEntity.value
+							)
+						} else {
+							await cacheHelper.entityTypes.delete(tenantCode, orgCode, modelName, originalEntity.value)
+						}
```
nevil-mathew: We should do the same for delete as well.

borkarsaish65: added the code.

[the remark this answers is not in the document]
borkarsaish65: @coderabbitai 

The comment is not valid for the following reasons:

displayProperties — Each org builds and caches its own display properties independently under org:{orgCode}:displayProperties. There is no fallback-write pattern where one org's display properties get cached under another org's key (see displayProperties.set() — it always writes to the caller's own org key). So _clearUserCachesForEntityTypeChange(organizationCode, ...) correctly clears only the affected org's display properties. No cross-org sweep is needed here.

mentee/mentor user caches — These caches store user profile values (e.g. "gender": "male"), not entity type definitions (labels, options, model mappings). A change to a default-org entity type definition does not corrupt profile data cached for users in other orgs. Their cached profiles remain valid.

The actual cross-org stale cache problem is already addressed — it only exists for the raw entityTypes:* entries, which are the ones cached under another org's key via the fallback read path. That is exactly what deleteEntityTypesAcrossAllOrgs handles with the wildcard tenant:{t}:org:*:entityTypes:... sweep. That is the only cache namespace where default-org data is written under a different org's key.

src/services/entity-type.js
nevil-mathew, line 1: The `_clearUserCachesForEntityTypeChange` call in `entity-type.js` (lines 131–136 for update and 370–375 for delete) currently clears user profile caches (mentee/mentor) only for a single organisation code.

We should extend this same logic so that when the default gets updated, profile caches are cleared across all organisations

nevil-mathew: Verify this `_clearUserCachesForEntityTypeChange` doesn’t accept 3rd parameter as an object.
It expects a string.

borkarsaish65: fixed now

borkarsaish65: fixed now

src/services/entity-type.js
```diff
@@ -122,7 +132,8 @@ module.exports = class EntityHelper {
 				orgCode,
 				tenantCode,
 				updatedEntity.model_names ? updatedEntity.model_names[0] : null,
-				updatedEntity.value
+				updatedEntity.value,
+				isDefaultOrg
```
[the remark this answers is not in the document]
borkarsaish65: yes this is the correct approach, currently we dont have strategy to idenitfies stale data via keys. 
Its better to evict all then to keep stale data for now.
@coderabbitai

src/database/queries/userExtension.js
```diff
@@ -712,11 +712,17 @@ module.exports = class MenteeExtensionQueries {
...
 	static async getAllUsersByOrgId(orgCodes, tenantCode) {
 		try {
+			const viewName = utils.getTenantViewName(tenantCode, MenteeExtension.tableName)
+
+			// null orgCodes = all orgs (used for default-org cross-org cache invalidation)
+			if (orgCodes === null) {
+				const query = `SELECT user_id FROM ${viewName} WHERE tenant_code = :tenantCode`
+				return await Sequelize.query(query, { type: QueryTypes.SELECT, replacements: { tenantCode } })
+			}
+
```
nevil-mathew: Why did we introduce this change now? Is there any place where this function is called without an org code? Why are we bringing in unnecessary changes?

src/services/entity-type.js
```diff
@@ -584,7 +620,10 @@ module.exports = class EntityHelper {
 					// Clear all mentor caches for this organization
 					try {
 						// Get all users who might be mentors in this organization
-						const users = await menteeExtensionQueries.getAllUsersByOrgId([organizationCode], tenantCode)
+						const users = await menteeExtensionQueries.getAllUsersByOrgId(
+							allOrgs ? null : [organizationCode],
+							tenantCode
+						)
 						const mentorUserIds = users.map((user) => user.user_id)
```
nevil-mathew: Is DB query needed here? Why can't we remove it by wildcard?

src/services/entity-type.js
nevil-mathew, line 1: Why doesn't deleteEntityTypesAndEntities have cache invalidation?

src/services/entity-type.js
borkarsaish65: deleting via wildcard now

src/database/queries/userExtension.js
borkarsaish65: removed this code now

src/services/entity-type.js
borkarsaish65: added the invalidation now.

borkarsaish65:
@coderabbitai review my latest changes in the PR

src/services/entity-type.js
```diff
@@ -311,12 +321,20 @@ module.exports = class EntityHelper {
...
+								entityToDelete.value
+							)
+						} else {
+							await cacheHelper.entityTypes.delete(
+								tenantCode,
+								organizationCode,
+								modelName,
+								entityToDelete.value
+							)
+						}
```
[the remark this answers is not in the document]
borkarsaish65: fixed now. @coderabbitai check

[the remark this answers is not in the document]
borkarsaish65: @coderabbitai verify again , I have tested. Check line 299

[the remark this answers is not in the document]
borkarsaish65: @coderabbitai did u check n verify ?

borkarsaish65:
@coderabbitai review my latest changes.

borkarsaish65:
@coderabbitai is my pr approved from your side ?

borkarsaish65:
@coderabbitai entityToDelete.model_names will never be empty or null. It is mandatory. 
I have done other change, review and verify.

borkarsaish65:
@coderabbitai review latest changes, approve if everything is okay.

borkarsaish65:
@coderabbitai just pushed new change, verify and approve if all good.

src/generics/cacheHelper.js
```diff
@@ -1558,6 +1587,14 @@ const displayProperties = {
 		await del(orgKey, { useInternal })
 		await del(tenantKey, { useInternal })
 	},
+	async deleteAll(tenantCode) {
+		// Delete all org-level displayProperties
+		const pattern = `tenant:${tenantCode}:org:*:displayProperties`
+		await scanAndDelete(pattern)
+		// Optional: delete tenant-level fallback
+		const tenantPattern = `tenant:${tenantCode}:org::displayProperties`
+		await scanAndDelete(tenantPattern)
```
nevil-mathew: Please remove this

borkarsaish65: done

PR #1634, opened 2026-04-24
Ticket 4512 fix; from a session where mentor is deleted was not able to be deleted by a Session Manager

borkarsaish65:
## Release Notes

- Skip mentor-extension validation when deleting a session (DELETE requests); mentor-extension lookup and permission checks are only performed for non-DELETE methods when mentor_id is present.
- When mentor extension is missing during non-DELETE updates, respond with MENTORS_NOT_FOUND instead of INVALID_PERMISSION.
- Session Managers can delete sessions even if the mentor extension is missing or would previously block deletion.

## Contribution Summary

|--------|------------:|--------------:|
| borkarsaish65 | 12 | 9 |

[Review Change Stack](https://app.coderabbit.ai/change-stack/ELEVATE-Project/mentoring/pull/1634)


borkarsaish65:
@coderabbitai full review

src/services/sessions.js
```diff
@@ -687,16 +687,19 @@ module.exports = class SessionsHelper {
...
+				let mentorExtension =
+					(await cacheHelper.mentor.getCacheOnly(tenantCode, userId)) ??
+					(await mentorExtensionQueries.getMentorExtension(userId, [], false, tenantCode))
+				if (!mentorExtension) {
+					return responses.failureResponse({
+						message: 'INVALID_PERMISSION',
+						statusCode: httpStatusCode.bad_request,
+						responseCode: 'CLIENT_ERROR',
+					})
+				}
```
nevil-mathew: First, confirm whether the permission check is actually required here. 

We can’t just ignore the methods. What happens if this is an update?

borkarsaish65: The original fix used method !== DELETE_METHOD which was too broad — it skipped the check entirely for DELETE but left the PATCH case broken when mentor_id is null on the session.

Updated the condition to:


if (method !== common.DELETE_METHOD && (sessionDetail.mentor_id || bodyData.mentor_id)) {
This makes the intent explicit — the mentor extension check only runs when there is actually a mentor to validate:

sessionDetail.mentor_id — covers the session manager update case, where userId is already overwritten with the assigned mentor's ID at line 684
bodyData.mentor_id — covers the mentor reassignment case, where userId is overwritten with the new mentor's ID at line 687
DELETE is always skipped since we're removing the session, not validating a mentor assignment
Scenarios verified:

Following cases verified:

Mentor updates/deletes own session ✓
Manager updates mentor's session ✓
Manager deletes mentor's session (mentor active or deactivated) ✓
Manager updates/deletes session where mentor_id is already null ✓
Reassigning to an invalid mentor via PATCH still returns INVALID_PERMISSION ✓

nevil-mathew: The current check is effectively acting as a mentor existence validation, not a permission validation.

Since `userId` has already been overwritten with `sessionDetail.mentor_id` or `bodyData.mentor_id` before this block runs, the check no longer validates the caller — it validates the target mentor being assigned. Actual authorization is already enforced earlier via `sessionDetail.created_by !== userId`.

Because of that, returning `INVALID_PERMISSION` here is misleading, especially for PATCH requests where the intent is just to reassign the session. A missing mentor should be treated as a validation error instead.

Restricting this validation to explicit mentor reassignment also avoids blocking:

* DELETE requests
* updates on sessions whose mentor was deleted/null historically
* unrelated PATCH updates

Suggested change:

```js
// Only validate when a new mentor is being assigned via PATCH
if (method !== common.DELETE_METHOD && bodyData.mentor_id) {
    const mentorExtension =
        (await cacheHelper.mentor.getCacheOnly(tenantCode, bodyData.mentor_id)) ??
        (await mentorExtensionQueries.getMentorExtension(bodyData.mentor_id, [], false, tenantCode))

    if (!mentorExtension) {
        return responses.failureResponse({
            message: 'MENTORS_NOT_FOUND',
            statusCode: httpStatusCode.bad_request,
            responseCode: 'CLIENT_ERROR',
        })
    }
}
```

This keeps the behavior aligned with the actual intent of the check and avoids conflating authorization with entity existence.

borkarsaish65:
@coderabbitai full review

PR #1646, opened 2026-07-16
added-ubuntu-setup-guide

MallanagoudaBiradar:


documentation/3.2.0/native/scripts/linux/create-databases.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-set -euo pipefail
```
rakeshSgr, line 2: @MallanagoudaB why this is removed

MallanagoudaBiradar: Hi @rakeshSgr  I removed it because the -e flag was causing the script to exit immediately if any of the psql commands returned a non-zero exit code

MallanagoudaBiradar: was not getting added 
when i refered my setup guide this line was not there

